PR #63, opened 2016-01-26
Allow setting an install prefix

phw:
With this patch one can specify an install prefix for installation like so:

```
./install.sh /usr/local
```

I have kept the default as `/usr`, but I actually think it should be changed to `/usr/local` as this is the common location for side installing software (and `/usr` is reserved for distribution packages).

This would also make it easier to build source packages as one can specify a completely different package root (e.g. `${pkgdir}` inside a PKGBUILD). But this is currently prevented by the root check. Not sure, I personally would just remove it. Or we only enable the check when the prefix is `/usr/...`.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,3 +1,11 @@
+if [ -z  "$1" ]; then
+   export PREFIX=/usr
+else
+   export PREFIX=$1
+fi
+
+echo "Installing to prefix ${PREFIX}"
+
 # Make sure only root can run our script
 if [ "$(id -u)" != "0" ]; then
    echo "This script must be run as root" 1>&2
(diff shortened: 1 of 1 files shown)


gnunn1:
Thanks for this. I use install.sh at development time so I can update all of the resources when I make a change and need to push it to my local file system to test it. Since I was managing packages as well, I used /usr to override the existing files there. There is another script in data/pkg that I use to create the build package, at  some point I should look at merging the two together as with your change it's pretty much doing the same thing now.

phw:
Thanks for merging. Now I see the createReleaseArchive.sh. Looks like this could just call install.sh with a proper path now for most of the work :)

dsboger:
`./install.sh $pkgdir/usr` works well for AUR, except I have to manually remove `gschemas.compiled`, which is a system-wide file.

phw:
> ./install.sh $pkgdir/usr works well for AUR, except I have to manually remove gschemas.compiled, which is a system-wide file.

Yes, in my own apps I solve this by making gschema compilation configurable via a flag. That way the install can compile the schema but when building packages this can be disabled. That is something to consider when expanding the install script.

PR #84, opened 2016-02-18
Prepare nautilus extension for localization

dsboger:
Hello, I took the liberty to add localization support in the nautilus extension (`open-terminix.py`) by using the `gettext` module. I used the same domain as the application to make it simpler, so I added 4 new keys to `terminix.pot`. Then I updated the pt_BR localization with the new keys.

### DIFF
--- a/data/nautilus/open-terminix.py
+++ b/data/nautilus/open-terminix.py
@@ -3,6 +3,10 @@
 import os
 import urllib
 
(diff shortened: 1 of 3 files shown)


frnogueira:
Nice!

gnunn1:
Thanks, really appreciate this.

PR #92, opened 2016-02-20
Localization issue fixes

phw:
This fixes a few issues with the localization, mainly missing strings and a syntax error in the pot files.

I also experimented with xgettext whether I could get the strings automatically extracted, but so far the results are not so great. I had more luck in the past with a project in a language unsupported by xgettext where I could just reuse the support for a syntactically similar language. I will try a bit more and let you know if I got something working.

Something else: Have you already thought about setting up some translation tool with git integration, something like Transifex or maybe Weblate? This could eliminate all the translation pull requests. Transifex is maybe the easiest one to setup and use, see e.g. https://www.transifex.com/phwolfer/Android_PwdHash/dashboard/ . Weblate is an open source web tool. It can be self-hosted, but they offer free online hosting for open source projects, too. I am using both tools, personally I would probably go for Transifex in this case due to the simplicity.

If you want I can help setting something up :)

### DIFF
--- a/po/terminix.pot
+++ b/po/terminix.pot
@@ -376,8 +376,8 @@ msgstr "Create a new session"
 msgid "Name"
 msgstr "Name"
 
-msgid "When command exists"
-msgstr "When command exists"
+msgid "When command exits"
+msgstr "When command exits"
 
 msgid "Could not locate dropped terminal"
 msgstr "Could not locate dropped terminal"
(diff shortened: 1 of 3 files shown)


frnogueira:
Great!

I will open a issue about this missing string. +1:

gnunn1:
Thanks for correcting those typos and adding the missing strings, much appreciated.

gnunn1:
Thanks for the suggestions on Transifex and Weblate, it's on my todo list
to make translations easier just been too busy with code to look at it in
much detail. I'd certainly appreciate the help in setting this up :)

PR #103, opened 2016-02-23
Add localization support for desktop file

dsboger:
This PR adds localization to the .desktop file. First two commits are required by msgfmt to work on desktop files. Last commit renames desktop file to ...desktop.in and adds a msgfmt command to install.sh that generates the localized desktop file. Also adds a new key to terminix.pot for the Comment property of the desktop file.


dsboger:
I've copy/pasted the headers from pt_BR.po and forgot to update the one in fr.po. Last commit fixes that.

PR #105, opened 2016-02-24
i18n extract strings

phw:
This is currently work in progress, but I'll appreciate feedback.

Trying to properly get the localization working with string extraction. I found out that using xgettext produces good results and is able to extract all strings from the D files when using `--language=Vala`.

The provided `extract-strings.sh` script will currently extract all strings for translation from the D source, the desktop file and the Nautilus extension. Some changes to the source were required, most notably the addition of a `N_()` function that allows to mark strings for translation without actually translating them. It also contains a couple of fixes to the strings itself.

**What's currently missing** is the extraction of the actions as defined in the GSettings file. I have no really good idea how to handle those. You can't use the GSettings language for xgettext as this will not do what we want here (it will just export the descriptions). We could use `N_()` on the relevant `ACTION_*` constants in the code, but they are all over the place and only some of them are indeed actions with keyboard shortcuts. Also there are things like `ACTION_SESSION_TERMINAL_X = "switch-to-terminal-"`.

Easiest way would probably be to just duplicate the relevant strings in a file just for extraction. Or actually generating such a file with a custom extraction from the GSettings file.


gnunn1:
Looks good, one question though. In appwindow.d and terminal.d you brought the '...' back into the menu items, won't that break existing translations since it changes the key?

Funny that Vala worked with xgettext, I had tried it with C originally and had no luck and gave up there. If you want some context to decisions I made you can read my blog entry here where I talk about localizing terminix: http://gexperts.com/wp/gtkd-and-localization/

The action strings are really only meant to be localized in one spot, the shortcut page of the preferences dialog. Otherwise they are considered data then anything else. In terms of extracting text for them, the only other idea I can think of is having a static associative array in code that maps the shortcut name to a proper string and then that string in turn is localizable. Basically similar to what is being done with encodings.

phw:
> Looks good, one question though. In appwindow.d and terminal.d you brought the '...' back into the menu items, won't that break existing translations since it changes the key?

Yeah, but a) will I take care of that when actually updating the pot file and b) does gettext handle that case pretty well. When merging the translations it will attempt to match e.g. `Load...` to existing translations, and it will find e.g `Load`. It reuses the existing `Load` translation for `Load...`, but marks it as fuzzy. The translator then just needs to check the fuzzy translation again and update it accordingly (removing the fuzzy flag). As long as a translation is marked fuzzy it is not used in the UI. That's a situation that will happen a lot when changing strings in the future ;)

> The action strings are really only meant to be localized in one spot, the shortcut page of the preferences dialog. Otherwise they are considered data then anything else. In terms of extracting text for them, the only other idea I can think of is having a static associative array in code that maps the shortcut name to a proper string and then that string in turn is localizable. Basically similar to what is being done with encodings.

I think I have a working solution, extracting the required strings with sed from the gschema file and using xgettext on the result. Maybe not the most beautiful way to handle it, but it avoids any duplication of the keys. Will update the pull request later.

phw:
I have updated the extraction script to collect the action names for keyboard shortcuts and to automatically merge new translations into the existing .po files. I have also run the script, fixed the changed source strings in existing translation and completed the German translation again.

From my point of view this is ready to be used.

gnunn1:
Thanks, I will review it in more detail tonite and merge.

gnunn1:
Thanks for all your hard work on this, greatly appreciated as I haven't done much localization with gettext myself.

PR #113, opened 2016-02-29
Improve install.sh portability

### LINKED ISSUE #112: Add shebang to install.sh
xenial$ sudo ./install.sh 
Installing to prefix /usr
./install.sh: 14: ./install.sh: function: not found
Processing 
msgfmt: no input file given
Try 'msgfmt --help' for more information.
./install.sh: 19: ./install.sh: Syntax error: "}" unexpected

This is because sudo tries to execute it via /bin/sh which is a symlink to dash on Ubuntu, whereas the script apparently needs bash. Please add a "#!/bin/bash" or "#!/usr/bin/env bash" line to the top.

dsboger:
I've made some changes to `install.sh` so that it works with `dash`, as a way to fix #112. I removed the `export -f` bashism and moved the po compilation commands to a `for` loop. I've also added a shebang pointing to `sh`. I've tested it with both `bash` and `dash` and it seems to work.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env sh
+
 if [ -z  "$1" ]; then
    export PREFIX=/usr
     # Make sure only root can run our script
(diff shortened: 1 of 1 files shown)


dejlek:
+1
I was thinking of doing the same!

PR #121, opened 2016-03-02
Updated translation files and German localization

phw:


frnogueira:
Starting to translations... ;)

PR #123, opened 2016-03-04
pt_BR.po 100%

frnogueira:


phw:
I don't think the binary .mo file should be included in the repository.

Also just because I struggled with this myself when doing the German translation: I think those key shortcut identifiers like `switch-to-session-0` are meant to replaced with a proper description. That's why https://github.com/gnunn1/terminix/blob/master/po/en.po#L94 redefines it with proper English "Switch To Session 0". You currently mimic the original technical identifier style "alternar-para-sessão-0"

frnogueira:
Sorry by that!

I forget to delete him... :(

When I translate the pt_BR.po, this file is empty... I did never look at en.po. ;)
I will change this...

Issue #136, opened 2016-03-14
Keybindings for directionnal terminal switch


noirbizarre:
Hi !

Previous and next terminal keybindings are sufficiant when you split only once but when you have, let's say 4 terms, it become difficult to go to the expected term.
It would be nice to have shortcut for:

- go to the left terminal
- go to the right terminal
- go to the above terminal
- go to the terminal below

gnunn1:
It's an interesting idea but no plans to do this, because of the way the Paned component component figuring out which is the right terminal based on direction is fairly complicated and I suspect error-prone.

alexwhitman:
Just for reference it's implemented in terminator [here](http://bazaar.launchpad.net/~gnome-terminator/terminator/trunk/view/head:/terminatorlib/window.py#L804)

It would be useful to have, I use it quite a bit in terminator.  Maybe it could be something for a plugin if plugins are on the roadmap?

gnunn1:
@alexwhitman Thanks, that's pretty much along the lines of what I'd expect the implementation to look like. I'll keep this feature request in mind in case I have a eureka moment but it still likes kind of error prone to me. I don't use Terminator so no experience with this feature, do you find it consistently makes the expected choice, i.e. no surprises in what terminal it selects for a direction?

No plans for a plugin capability at this time, since Terminix is open source and people can just submit pull requests for now.

alexwhitman:
> do you find it consistently makes the expected choice, i.e. no surprises in what terminal it selects for a direction?

That tends to depend on the layout but I think it can be improved.  For example, if you vertically split once, then in the right hand side horizontally split so you have:
```
|-----|-----|
|     |  2  |
|  1  |-----|
|     |  3  |
|-----|-----|
```
Moving right from 1 ends focuses 3 but I would personally expect it to focus 2.

I've never used D before but if I get some time I might take a look at implementing it.

phw:
That's actually the one feature I miss most in Terminix which prevents me to completely abandon Terminator. I find myself using this way of navigating between the terminals all the time.

gnunn1:
@alexwhitman From a design perspective, I would think a simple brute force approach would actually work best over trying to do anything to fancy. Using your example layout, I'd look at driving the direction off of the caret position in the focused terminal and then start sampling every 20-40 pixels or so in the desired direction until you hit a different terminal or are out of screen. Using that approach if the caret is at the top of 1, 2 would be selected. If the caret was at the bottom 3 would be selected.

[...]

PR #157, opened 2016-03-19
Add directional terminal switching

alexwhitman:
Implements #136. This adds directional terminal switching using Alt+Up/Down/Left/Right.

The "next" terminal is determined by moving from the top left coordinates of the currently focused terminal rather than the position of the cursor.

### DIFF
--- a/data/gsettings/com.gexperts.Terminix.gschema.xml
+++ b/data/gsettings/com.gexperts.Terminix.gschema.xml
@@ -514,6 +514,22 @@
       <default>'&lt;Alt&gt;9'</default>
       <summary>Keyboard shortcut to switch to terminal 9</summary>
     </key>
+    <key name="session-switch-to-terminal-up" type="s">
+      <default>'&lt;Alt&gt;Up'</default>
+      <summary>Keyboard shortcut to switch to terminal above</summary>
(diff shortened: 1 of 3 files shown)


gnunn1:
@alexwhitman Thanks, I had a quick initial look and it looks fine to me. When I have more time tomorrow I will test it out and merge it in. Only change I think it needs is to check if the terminal is maximized by checking maximizedInfo.isMaximized and ignoring the request for the focus change. I would suggest focusing the terminal by calling the session.focusTerminal(terminalID) instead of calling terminal.focusTerminal directly. I'm trying to centralize my focus checks in that one method. If you don't have time I can take care of it tomorrow.

Really appreciate you taking the time to put this together.

alexwhitman:
I've pushed a new version which uses `session.focusTerminal(terminalUUID)`.

`session.focusTerminal(int)` already checks `maximizedInfo.isMaximized` but `session.focusTerminal(string)` doesn't.  The two could be consolidated so that the maximized check is just in one place but I'll leave that up to you to decide.

gnunn1:
Yeah I was just going to reply to ignore my request as I was going to re-work this, thanks again.

PR #175, opened 2016-03-23
Fixed shortcut pref disabled string

phw:
This fixes actually two issues:

1. When using Terminix with a non English locale, the localized string for "disabled" got saved to the preferences. This lead to errors while loading the preferences. This PR fixes this by not localizing the value actually saved.  Introduced by myself in bd75e68db72cdd5864b13923db1de2815e114e33, sorry for that.

2. The disabled string was not displayed anymore.


gnunn1:
Thanks for the fix

PR #177, opened 2016-03-23
Add split buttons to popover menu

alexwhitman:
Adds the split buttons to the right click popover menu.


gnunn1:
Thanks again @alexwhitman, much appreciated

PR #181, opened 2016-03-25
Fix sidebar transparent background

alexwhitman:
As of the current master branch the sidebar background on Ubuntu 16.04 is transparent:
session-switcher

This pull request adds a background colour to the sidebar.


gnunn1:
I merged this PR but changed the color to the theme_bg_color rather then theme_fg_color since this was the behavior everywhere except Ubuntu. Ubuntu really needs to update that Ambiance theme at some point.

Thanks for taking care of this @alexwhitman

PR #183, opened 2016-03-26
Add double click in sidebar to open new session

alexwhitman:
With this change, double clicking in the empty space of the sidebar will open a new session.

Wasn't sure if there is a better way to get the AppWindow instance to be able to call `createSession` but if there is I can make the change.

### DIFF
--- a/source/gx/terminix/appwindow.d
+++ b/source/gx/terminix/appwindow.d
@@ -147,6 +147,7 @@ private:
         }, ConnectFlags.AFTER);
 
         sb = new SideBar();
+        sb.setAppWindow(this);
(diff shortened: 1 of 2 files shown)


gnunn1:
Generally I've been using events to try to avoid coupling certain objects together.  Having said that, I'm not very keen on this change, do any other Gnome or well known applications behave this way? I tried the gedit side panel and it doesn't do this.

alexwhitman:
Browsers would allow double clicking in the tab bar to open new tabs but that seems to have changed recently.

I only really implemented it so that it was a bit quicker to open a new session after seeing the session list but looking back I agree that it's not a great solution.

PR #203, opened 2016-04-03
po/zh_CN: filled in

MingcongBai:


MingcongBai:
BTW, just as a random note... Can we move some of these work onto Transifex so it will be easier to notify translators (regarding changes in the translation), and for you to import them when time comes for a release (and make them do it quicker before it would be too late).

gnunn1:
Absolutely, it's been suggested before just haven't had time. Can you open an issue for it so I don't lose track of it.

PR #226, opened 2016-04-11
Shortcuts help: Mark section titles as translatable.

phw:
This makes the section titles in the shortcut overview translatable (the dropdown in the header bar).

To actually take effect this requires an update of the pot/po files.


gnunn1:
Thanks for catching this

PR #227, opened 2016-04-11
Added .desktop and terminix binary to gitignore

phw:


phw:
There is an error in my PR. It was meant to replace `./terminix` with `/terminix`, but I also removed the slash by accident. That makes it also ignore everything under `src/terminix`. Sorry for that.

PR #235, opened 2016-04-12
Fixed split terminal action names in shortcut help

phw:


gnunn1:
Thanks @phw, you are on a roll. As a heads up, i need to fix those actions in the On New Instance preference tonite as well, might be a localization change there in terminology as well from Horizontal/Vertical to Right/Down. I really should have thought this through more :)

PR #255, opened 2016-04-16
Fix link OpenSUSE Build Service

dotws:


gnunn1:
I'm rejecting this PR because I realize I typo'ed the OBS version number and that's why the URL changed, your PR showed me what the problem was though so thanks for that! I've kicked off a new build in OBS so hopefully it will be corrected shortly.

PR #257, opened 2016-04-17
Create appdata folder in install.sh

dsboger:
The current installer copies the terminix appdata file to a file named appdata. Fixed creating appdata folder prior to that.


dsboger:
Also, please fix the official release package.

PR #278, opened 2016-04-26
Adding terminal reset & clear features

Boiethios:
I added some reset & clear features, like in gnome-terminal. I never used D, so if I have made something wrong, I'd like to know what is it ;)

It seems to work. However, I could not find how to add the new entries in the translation files.


gnunn1:
Thanks, good effort but I'd need some changes made before I could accept:

a. Change the names of the actions from 'reset-terminal' and 'clear-terminal' to just 'reset' and 'clear' as they are already scoped to the terminal via the ACTION_PREFIX. The constants should be re-name to ACTION_RESET and ACTION_CLEAR as well.

b. You need to add shortcuts for these to com.gexperts.Terminix.gschema.xml and register the actions using the method registerActionWithSettings. The shortcut names in the xml file should be terminal-reset and terminal-clear

c. I notice both actions just call vte.reset, have you checked the source code to see what gnome-terminal is doing. Also, I think in gnome-terminal it is ```reset and clear``` not just ```reset```?

Boiethios:
Ok, I will make it before another request.

a. I did not know. This is simpler.
b. I'll do it.
c. The names are not accurate enough, indeed. The same method is called in both cases, once with true, once with false. The `reset` option reinitialize the terminal states ; the `reset and clear` options do the same **and** clear the output buffer (very useful when an unlimited output is set). I'll change the names.

Boiethios:
That is done.

gnunn1:
Thanks, I'll do a merge when I get home from work tonite. One change I might make though as an FYI is remove the menus and just leave it accessible via shortcuts, I'll need to see how big the menu is looking particularly the context menu when the terminal titlebar is hidden.

phw:
And don't forget the shortcuts should also be added to the help overlay :)

I somewhat dislike the fact that the help overlay duplicates a lot of information already present elsewhere (shortcut info in schema, descriptions in the configuration screen) :(

Boiethios:
Perhaps a submenu `Terminal` with `Read-Only`, `Reset`, `Reset and Clear` could be a good solution ? And it can be populate by some other stuff in the future, when needed.

gnunn1:
@phw Yep, I was planning on taking care of that myself just because you have to hand edit that UI file since glade doesn't support Shortcut widgets yet.

I don't disagree with you, I had actually planned on building the shortcut window from code based on the schema file but it looked like crap, it really needed one more level of categorization. If I had to re-design it again, I'd probably change all the shortcut names in the schema file to be prefix-category-actionname where category would provide the second level of nesting.

gnunn1:
@Boiethios That might work, it's difficult for me to say for sure without seeing it though.

Boiethios:
I have done it, I think it is much better.

PR #306, opened 2016-05-07
Fixed the .desktop categories

Mailaender:
The choice of categories was not adhering to https://freedesktop.org/wiki/Specifications/menu-spec/ which is punished by a failing build in @openSUSE. This should fix it.


gnunn1:
Looking at openSUSE documentation it seems like Utility is supported as a category and as far as I can tell seems quite commonly used by different applications, what am I missing?

https://en.opensuse.org/openSUSE:Packaging_desktop_menu_categories

Mailaender:
Yes, but you can't just throw in a main category. It has to also define a sub-category like `Utility;Shell;`. However note that terminal emulators shouldn't use that one, but `System;TerminalEmulator;` instead.

PR #312, opened 2016-05-09
Do not remove default profile when adding new one

### LINKED ISSUE #294: Added a new profile, existing profile dissapeared
This is a bit strange, because I cannot reproduce it, but it definitely happened. I had only one default profile and all my terminals and saved sessions were configured to use it. I went in settings, added a second profile and then wanted to switch some open terminals to use this profile. I then realized that the old profile was missing in the list, the existing terminals had no profile selected. I can tell that I did not edit the existing profile by accident since the ID changed and it was unselected for the existing terminals.

I just wanted to report this now, but this is probably difficult to act upon until somebody is able to reproduce it. I'll try.

phw:
Fixes #294

### DIFF
--- a/source/gx/terminix/preferences.d
+++ b/source/gx/terminix/preferences.d
@@ -203,11 +203,6 @@ public:
         string[] ps = gsProfileList.getStrv(SETTINGS_PROFILE_LIST_KEY);
         trace("Get list of profiles");
 
-        //Remove default if it is there, only happens on first run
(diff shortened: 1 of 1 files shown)


gnunn1:
I think it might be there because early in Terminix development I may have been explicitly creating a profile. Now that you've narrowed down what the problem is, I'll double-check it and if everything looks good merge your change.

gnunn1:
Fix #294

Issue #301, opened 2016-05-05
Gettext used fundamentally incorrectly for shortcuts


egmontkob:
At certain places, Terminix uses internal technical identifiers as the string to be translated (msgid), and translates it even to English. E.g. en.po contains:

msgid "switch-to-terminal-1"
msgstr "Switch To Terminal 1"

This approach is wrong for a couple of reasons.

Sure there are some internationalization frameworks out in the wild that work this way, but gettext is not one. It expects the source string to be the American English UI label. (In case the English word is ambiguous and the same word appears at multiple places which should be translated differently, the solution is pgettext()).

Gettext falls back to the source string (msgid) and displays that on the UI if the catalog(s) according to the LANG, LANGUAGE, LC_* variables do not contain the translation.

Just start up Terminix with a locale that it is not translated to. E.g. `LC_ALL=hu_HU.UTF-8 LANGUAGE= terminix`. The UI literally presents `switch-to-terminal-1`. Not good.

Also, translators are not being helped enough. E.g. they see `switch-to-terminal-1`, it's hard to come up with a consistent way of capitalizing the letters – how should they know, without seeing the English translation, whether it's preferably translated with fully lowercase, or uppercase initial, or uppercase initial at each word? Sure a language might decide to deviate from English, but at least there should be a reference. Even flavors of English, let's say Canadian, might easily get "translated" differently from the American one for no good reason.

Sometimes the ID is even less intuitive, e.g. `switch-to-terminal-0` becomes `Switch To Terminal 10`, a translator shouldn't have to figure out that 0 stands for 10.

`app` is translated to `Application`, `win` is translated as `Window` (a translator might believe it's the opposite of `lose`) etc.

Please stick to the convention and how gettext was designed: msgid should always be the American English UI string. en.po probably shouldn't exist since the American strings are hardcoded; if other spellings are available then they go to en_GB.po, en_CA.po etc.

gnunn1:
Well I am Canadian...

Seriously though, I'm not adverse to a better way of doing it but I'm not seeing an obvious solution that wouldn't have its own set of problems. For example, I could create a source file that has a static hash map that maps shortcuts to English names and localize that. On the downside though it's yet another file that needs to be updated and maintained when shortcuts are added or changed. I already have gsettings and shortcuts.UI to maintain, I'd prefer not to add a third file.

[...]

PR #313, opened 2016-05-10
Fix reading the shortcut localization with correct context.

phw:
This fixes the shortcut labels to be localized with the proper localization context defined inside `shortcuts.ui` and finalizes the changes introduced for #301 .

Another solution could have been to remove the localization context, but I saw that some of the strings are also used as menu labels, and it is actually useful to be able to have a different (maybe longer) translation in the help text. So I opted for keeping the context but make sure it gets used.

I named the function `C_()` to follow the naming of the C macro commonly used for this.

On a side note the file saving workaround in `l10n.d`can probably be removed. Since it is only used in `_()` and not in the `N_()` and now `C_()` functions it isn't working properly anyway.

### DIFF
--- a/extract-strings.sh
+++ b/extract-strings.sh
@@ -13,6 +13,7 @@ find ${BASEDIR}/source -name '*.d' | xgettext \
   --files-from=- \
   --directory=$BASEDIR \
   --language=Vala \
+  --keyword=C_:1c,2 \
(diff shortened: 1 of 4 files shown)


gnunn1:
Thanks @phw, really appreciate all your help with this as the localization support in Terminix is largely due to your efforts.

PR #318, opened 2016-05-10
Localize buttons in FileChooserDialog

phw:
Gtk-D will otherwise use hardcoded defaults that are not translated, see https://github.com/gtkd-developers/GtkD/blob/master/src/gtk/FileChooserDialog.d


phw:
I think with this we finally get a 100% localizable Terminix :D

gnunn1:
Awesome, great work

PR #324, opened 2016-05-13
Add experimental autotools support

dsboger:
This PR adds the necessary infrastructure to build terminix with autotools. Some documentation follows (I hope I did not forget anything).

**Requirements**
The following assumptions are made about the system:
- autoconf >= 2.69, automake >= 1.15 and gettext >= 0.19.7 are installed (may work with older versions, but not tested by me)
- dmd compiler executable is in the $PATH
- pkg-config is installed
- gtkd-3 and vted-3 (or gtkd-3-dmd and vted-3-dmd) shared libraries and "headers" are installed and were compiled with dmd (instructions to build these below)

**Building**
In order to build terminix with autotools, issue the following commands:
```sh
./autogen.sh
./configure
make
sudo make install
```
- `./configure` accepts a `--prefix=/path/to/prefix` parameter (defaults to `/usr/local`)
- The compiler flags are passed to dmd through variable `DCFLAGS`. Default value is `-O`. A release version would be built with 

```make DCFLAGS='-O -inline -release -version=StdLoggerDisableTrace'```

- One may use `sudo make DESTDIR=/path/to/install install` to do a "staged install" (used mostly by packagers)

It is possible to create a source distribution tarball by calling `make dist` (after configure step). This will create a file `terminix-<version>.tar.gz` that can be used to build terminix without autotools, by extracting the contents and running well known `./configure; make; make install` commands inside the extracted directory.

**Maintenance**
If no new files are added to the tree, maintaining autotools files means simply bumping versions in `configure.ac`. The following file additions also do not require any additional maintenance:

- `.d` files under `source/`
- `.json` color schemes under `data/schemes/`
- `.css` stylesheets under `data/resources/css/`
- `.ui` GtkBuilder files under `data/resources/ui/`
- icons under `data/resources/icons/`
- icons under `data/icons`

These additional files require some actions:

- new `.po` language files under `po/`: the respective language code must be added to `po/LINGUAS` or the respective translations will not be installed
- other files must be analyzed on a case-by-case basis

**Building GtkD shared libraries**
GtkD tarball comes with a handmade makefile, so building and installing it is very easy:

```
make DC='dmd' DCFLAGS='-O -release -inline' shared-libs shared-gstreamer shared-vte
make prefix=/path/to/prefix DESTDIR=/path/to/install install-shared install-shared-gstreamer install-shared-vte install-headers install-headers-gstreamer install-headers-vte
```

These commands install all the shared libraries, although terminix only requires the base and vte ones (headers for these are also required).


gnunn1:
@dsboger Thanks for this, can you update the wiki with these instructions and then in the Readme add a note in the build section about experimental autotools support with a link to the section in the wiki.

I'll do the same for the xdg-app stuff that mclasen checked in as well.

dsboger:
I'll do that, but it's the first time I edit a GitHub Wiki. Do I need some special permission from you to edit it online or should I clone it, work offline and propose some kind of PR? The GitHub help does not mention community contributions to wikis and my fork does not contain the upstream wiki, so I'm kind of lost.

gnunn1:
@dsboger Sorry for the delay in replying, I've opened up the wiki editing to allow anyone to edit the wiki.

dsboger:
I've created a Wiki page with a text similar to the above.

PR #354, opened 2016-05-29
Update Russian localization

igor-dyatlov:
Updated translation


gnunn1:
In the future, please do translations in weblate as per the project front page.

PR #355, opened 2016-05-29
add travis support

wilzbach:
It's pretty easy to automatically test for various compilers.
See:

https://travis-ci.org/wilzbach/terminix

You need to enable Travis after this is merged.
It's currently failing because of one line

```
source/gx/terminix/prefwindow.d(314): Error: no property 'clear' for type 'string[string]'
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: d
+d:
+ - dmd-2.071.0
+ - dmd-2.070.2
+ - ldc-1.0.0-beta2
+script:
+ - dub build


gnunn1:
That failure is due to not having the latest and greatest phobos as they recently added the ability to clear associative arrays. Is that the only issue for the ldc beta, if so I'm fine with putting in a workaround for that line.

wilzbach:
Afaik it's in druntime - there is also a ldc branch with 2.071 that compiles flawlessly,  but it probably takes a while until this is published. 
Please note that ldc 1.0.0 isn't available on debian as they have their testing, unstable etc model. 
In any case I guess there also seem other problems with packaging on debian, so hopefully once they are resolved  it just works :)

wilzbach:
> . Is that the only issue for the ldc beta

Yes and as you can see it produces the same error with dmd 2.070

gnunn1:
Thanks, I've removed the clear and just the associative array to null to clear it.

wilzbach:
Are you familiar with travis?
You can now simply enable the project, s.t. travis checks all future commits & PRs.

https://docs.travis-ci.com/user/getting-started/
Enable here: https://travis-ci.org/profile

gnunn1:
No I'm not familiar with Travis though I have experience with CI tools in the Java world like Jenkins/Hudson. I'll have a look later tonite when I get the chance, I'm a big believer in CI in general so I'm all for getting this going.

gnunn1:
My apologies for the delay, I added terminix to Travis and it can be viewed here:

https://travis-ci.org/gnunn1/terminix

PR #379, opened 2016-06-19
using env so that it will run on more systems

jonorthwash:


gnunn1:
I'm curious, are there any Linux distros that can't work with that shebang, I always thought the environment thing was mostly useful when trying to use the same script for other nix's like BSD?

jonorthwash:
Didn't work on my system: for me it's `/bin/sh`.  Using `env` is safer, pretty sure.

gnunn1:
What distro are you using?

jonorthwash:
Current debian unstable.

PR #391, opened 2016-07-13
i18n: Localize the profile switch description.

phw:
Make sure the window can automatically break the text so it adapts well to different translations.


phw:
You have tr re-run extract-strings after this, best after also merging #392

PR #393, opened 2016-07-13
Improve metainfo file

ximion:
Adds more metadata and removes superfluous or outdated attributes/paths.

This kills some hints from the AppStream metadata generator. Please note that changing the install directory requires up-to-date AppStream tools.
At time, the current (development) releases of all major distributions support this, including Arch, Debian, Ubuntu, Fedora, OpenSUSE. Changing the path might affect backports for some distros (Debian, Ubuntu and Arch are safe, so is OpenSUSE, I am not 100% sure about Fedora).

See https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps for more information on how to write good metainfo files for apps.

### DIFF
--- a/data/appdata/Makefile.am
+++ b/data/appdata/Makefile.am
@@ -1,4 +1,4 @@
-appdatadir = $(datadir)/appdata/
+appdatadir = $(datadir)/metainfo/
 appdata_DATA = com.gexperts.Terminix.appdata.xml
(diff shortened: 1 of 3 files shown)


dsboger:
This is changing the location where the appdata file is installed. Will it be still found by the appstream tools?

PR #409, opened 2016-07-21
Focus terminal on middle click paste

### LINKED ISSUE #408: Focus the session on pasting with scrollwheel button
Currently, when user tries to paste a text to a session which is out of focus using a scroll wheel button, the paste works fine, but the sessions stays out of focus.

It would be nice if it got focused just like when user clicks to it

alexwhitman:
Fixes #408

### DIFF
--- a/source/gx/terminix/terminal/terminal.d
+++ b/source/gx/terminix/terminal/terminal.d
@@ -1174,6 +1174,9 @@ private:
                 pmContext.setPointingTo(&rect);
                 pmContext.showAll();
                 return true;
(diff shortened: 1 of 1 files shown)


alexwhitman:
Just noticed while testing this, middle click pastes are only sent to a single terminal even when synchronized input is enabled.

gnunn1:
Thanks Alex for the quick fix. With regards to the middle-click paste, yes that's a bug so good catch. I think I need to handle the addOnPasteClipboard from vte and move the synchronization code for handling pastes from the action handler to this.

PR #418, opened 2016-07-26
Overwrite file on save, fixes #417

### LINKED ISSUE #417: cannot save output to a existing file
When I save output to a existing file, terminix will exit with the following log:
```
2016-07-26T15:54:23.295:app.d:main:99 Unexpected exception occurred
2016-07-26T15:54:23.295:app.d:main:100 Error: Error opening file '/home/tiancj/Desktop/output.txt': File exists

```
the version of terminix is 1.2.0.

alexwhitman:
`create()` will throw an error if the file already exists.  `replace()` will overwrite the file, or create it if it doesn't already exist.

Fixes #417

### DIFF
--- a/source/gx/terminix/terminal/terminal.d
+++ b/source/gx/terminix/terminal/terminal.d
@@ -1953,7 +1953,7 @@ private:
         }
         //Do work here
         GFileIF file = GFile.parseName(outputFilename);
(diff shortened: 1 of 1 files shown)


gnunn1:
Thanks @alexwhitman

Issue #430, opened 2016-08-05
Installer errors


alexwhitman:
Since 417dc23fbaa8eedb146d6cafea8b4f808a6bb5a2 has been merged I'm now getting installer errors.

Firstly it was
```
Creating PNG symbolic resource icons
./install.sh: 37: ./install.sh: gtk-encode-symbolic-svg: not found
./install.sh: 38: ./install.sh: gtk-encode-symbolic-svg: not found
./install.sh: 39: ./install.sh: gtk-encode-symbolic-svg: not found
./install.sh: 40: ./install.sh: gtk-encode-symbolic-svg: not found
Building and copy resources...
terminix.gresource.xml: Failed to locate 'icons/16x16/actions/terminix-add-horizontal-symbolic.symbolic.png' in current directory.
```
which was solved by installing `gtk-3-examples` on Ubuntu.
After that the error I'm getting now is
```
Creating PNG symbolic resource icons
Can't save file icons/16x16/actions/terminix-add-horizontal-symbolic.symbolic.png: Error opening file '/home/alex/projects/github/terminix/data/resources/icons/16x16/actions/terminix-add-horizontal-symbolic.symbolic.png': No such file or directory
Can't save file icons/16x16/actions/terminix-add-vertical-symbolic.symbolic.png: Error opening file '/home/alex/projects/github/terminix/data/resources/icons/16x16/actions/terminix-add-vertical-symbolic.symbolic.png': No such file or directory
Can't save file icons/32x32/actions/terminix-add-horizontal-symbolic.symbolic.png: Error opening file '/home/alex/projects/github/terminix/data/resources/icons/32x32/actions/terminix-add-horizontal-symbolic.symbolic.png': No such file or directory
Can't save file icons/32x32/actions/terminix-add-vertical-symbolic.symbolic.png: Error opening file '/home/alex/projects/github/terminix/data/resources/icons/32x32/actions/terminix-add-vertical-symbolic.symbolic.png': No such file or directory
Building and copy resources...
terminix.gresource.xml: Failed to locate 'icons/16x16/actions/terminix-add-horizontal-symbolic.symbolic.png' in current directory.
```

gnunn1:
I'm not a big fan of having generated artifacts in source control but in this case I think I'll make an exception and simply checkin the symbolic png icons. Forcing people to install the examples package to build terminix doesn't seem like a good idea?

Thoughts from @alexwhitman and @phw?

gnunn1:
[...]

PR #431, opened 2016-08-05
Fix svg resource install

phw:
See issue #430


gnunn1:
I've opted to check-in the icons so debian doesn't require the gtk-3-examples package to create the symbolic icons.

PR #449, opened 2016-08-18
Add support for building with Meson

ximion:
Hi!
I am not sure if you actually want this patch, because Terminix has two build systems already, but since I wrote it already, making a PR out of it and ask for feedback won't hurt.

So, what is Meson and why should you use it?
Meson is a fast, Ninja-based buildsystem where I recently added D support to (currently under review at https://github.com/mesonbuild/meson/pull/685). It is much more powerful than dub and also compiles much faster than dub.
In spirit, it is comparable to CMake/Autotools, but without all the craziness and following a very clean and simple design.

See https://ninja-build.org/ for information on Ninja (which also compiles stuff like e.g. Google Chrome) and http://mesonbuild.com/ for Meson.

To give a few numbers, a compilation of Terminix using dub takes about 8.6s on my system, while compiling it with Ninja takes only roughly 6s (the difference becomes really huge if you have a lot of D sources). This is using the LDC D compiler, creating a non-optimized debug build.

Meson is seen as one of the potential replacements of Autotools in GNOME and other projects.
I think, when D support is upstreamed, Terminix could very well drop the Autotools support and be Meson only - but that's up to you and whoever maintains the Autotools stuff right now.
(I myself prefer it over dub, even)

To test this, you can Git-clone my Meson fork[1], install Ninja from your distributor, and then run:
```
mkdir b && cd b
./path/to/meson.py ..
ninja
```
[1]: https://github.com/ximion/meson
Missing in the implementation is currently support for downloading GtkD from Github or DUB, so this relies on GtkD being installed system-wide. This auto-download support with preferring system copies can be added to the build definitions later though, via Meson subprojects and wrapfiles (which is pretty cool).
So far, I primarily used the Terminix case as one of my more complex tests for Meson's D support.

I hope you find this useful!
Cheers,
    Matthias

### DIFF
--- a/data/icons/meson.build
+++ b/data/icons/meson.build
@@ -1,0 +1,14 @@
+
+icon_sizes = ['16x16', '22x22', '24x24', '32x32', '48x48', '256x256', '512x512']
+
+# install bitmap icons
+foreach size : icon_sizes
+    install_data('hicolor/' + size + '/apps/com.gexperts.Terminix.png',
+                 install_dir: 'share/icons/hicolor/' + size + '/apps/'
+    )
+endforeach
+
+# install scalable symbolic icon
+install_data('hicolor/scalable/apps/com.gexperts.Terminix-symbolic.svg',
+             install_dir: 'share/icons/hicolor/scalable/apps/'
+)
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,164 @@
+project('Terminix', 'd')
+
+if meson.get_compiler('d').get_id() == 'gcc'
+    error('Terminix can not be compiled with GDC at time, since the standard library version of all GDC versions is currently (2016) too old. Sorry.')
+endif
+
+glib = import('gnome')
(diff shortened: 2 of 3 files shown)


gnunn1:
I don't know have much interest in maintaining it, as you said two build systems is enough. However, I think there is merit in keeping your code around and I've been thinking about creating an experimental folder for things that aren't actively maintained but worth keeping around. 

An example is the flatpack stuff that mclasen did, I really want to move that out of the generic build folder it's currently in so was thinking experimental/flatpack. We could do the same with this if that makes sense to you?

ximion:
Doesn't sound like a bad idea.
IMHO it's worth considering to replace the Autotools stuff in Terminix with this patch though. The Meson build system does all we want in distributions, and it is way easier to maintain that Autotools (it also supports translations ^^).
For that to happen, D support must be in Meson though (which will happen, upstream is in favor of it and I have been working with them on this).

Issue #454, opened 2016-08-22
Quick session opener/switcher


alexwhitman:
Incoming feature suggestion...

Sublime Text has a [quick open panel](http://docs.sublimetext.info/en/latest/file_management/file_management.html#goto-anything) which performs a fuzzy search on files.

I think something similar for Terminix which searches for saved session files could be useful so that I could hit a hotkey, type something like 'web prod' (which would be in the name of a previously saved session file) and have the session open.

gnunn1:
Interesting but might be tricky to implement. For one, you'd likely one to have a preference of where to search for these files so you are not searching the entire home directory. The other issue wouldn't you'd have to actually read the file to determine if it's a terminix json file or some other json file?

alexwhitman:
The way Sublime Text does it is to store a list of paths to project files that it knows about. This list could be updated when a session file is opened or saved. Essentially it would be like a recent files list, but one which is fuzzy searchable.

alexwhitman:
This is something along the lines of how I think it could work:
```diff
diff --git a/source/gx/terminix/appwindow.d b/source/gx/terminix/appwindow.d
index 3952dcf..5a4cb33 100644
--- a/source/gx/terminix/appwindow.d
+++ b/source/gx/terminix/appwindow.d
@@ -141,6 +141,8 @@ private:
     // Track size changes, only invalidate if size really changed
     int lastWidth, lastHeight;
 
+    string[] recentSessionFiles;
+
     /**
      * Create the user interface
      */
@@ -839,6 +841,7 @@ private:
                 throw new SessionCreationException("Session could not be created due to error: " ~ e.msg, e);
             }
         }
+        addRecentSessionFile(filename);
         trace(format("Session dimensions: w=%d, h=%d", width, height));
         Session session = new Session(value, filename, width, height, nb.getNPages() == 0);
         addSession(session);
@@ -861,9 +864,11 @@ private:
         if (fcd.run() == ResponseType.OK) {
             try {
                 loadSession(fcd.getFilename());
+                addRecentSessionFile(fcd.getFilename());
             }
             catch (Exception e) {
                 fcd.hide();
+                removeRecentSessionFile(fcd.getFilename());
                 error(e);
                 showErrorDialog(this, _("Could not load session due to unexpected error.") ~ "\n" ~ e.msg, _("Error Loading Session"));
             }
@@ -900,6 +905,7 @@ private:
                 return;
             }
         }
+        addRecentSessionFile(filename);
[...]

PR #456, opened 2016-08-24
[WIP] Quick session loader (starts to implement #454)

alexwhitman:
Thought I'd put up some work in progress. just to check that I'm on the right lines.  The first commit implements the recent session file tracking.  The second commit starts the UI.

If you save a few session files and then hit `Ctrl+Alt+p` the UI will pop up.

Lots of things left to do:
- [x] Filter list when entering text into the search box
- [x] Open session file when hitting Enter in the list
- [x] Focus main window when hitting Escape in the search box
- [x] Handle removing entries from the recent files list when hitting Delete on a focus list box entry
- [x] Make the thing look a lot nicer
- [ ] Other tweaks and fixes I haven't thought about

Any guidance or contributions greatly welcome. :smile:


gnunn1:
I had a quick look at the code and ran it locally, I think you are on the right path but as you correctly noted there is quite a bit of work to do. A couple of comments as follows:

a. It doesn't close automatically when you click outside of it
b. You'll need to update the.ui file with the new shortcut at some point
c. Should we be keeping track of the session name as well and searching by that rather then filename?

alexwhitman:
I've pushed another commit which handles opening a session file and removing entries from the list.

> a. It doesn't close automatically when you click outside of it

Yes, there's a load of focus related issues which need to be sorted out, at this point I'm not entirely sure what I'm missing but I haven't really looked at it yet.

> b. You'll need to update the.ui file with the new shortcut at some point

That was already done in the second commit.

> c. Should we be keeping track of the session name as well and searching by that rather then filename?

Do-able, although I'm not sure how often sessions get renamed.  I have quite a few session files but I haven't changed them and they all have the name "Default".

In thinking about how to filter the list box when entering search terms into the search box, it seems as though `ListBox::setFilterFunc` would do what I want but the signature for that is
```
void setFilterFunc (GtkListBoxFilterFunc filterFunc, void* userData, GDestroyNotify destroy); 
```
and I can't find any useful information on how to use it within GTKD, particularly the `GtkListBoxFilterFunc`.

gnunn1:
The filterfunc that you set is declared as:

extern(C) int function(gtkc.gtktypes.GtkListBoxRow*, void*) filterFunc

So it looks pretty straight-forward, going by the docs it looks like it would be called for each row and I assume you would return a 0/1 to indicate whether the row should be shown or not.

BTW, I ideally recommend building the ddox version of the GtkD documentation locally, it's light years better then the on-line docs and makes understanding GtkD mich easier. If you need any help with ddox just ping me off-line (gerald dot b dot nunn at gmail dot com).

alexwhitman:
This might be a stupid question but what's the D method of writing a function to implement a GtkListBoxFilterFunc?

gnunn1:
My C-foo is weak, but I think it would be something like (assuming it's in a class not a standalone function otherwise remove the static)

extern (C) static int filterCallback((gtkc.gtktypes.GtkListBoxRow* prow, void* userData) {
   ListBoxRow row = ObjectG.getDObject!(ListBoxRow)(prow);
   if (row !is null) {

   }
}

alexwhitman:
Managed to figure out the filtering bit after a while and many casts between data types.

Right now it only does a basic string in string filter rather than a fuzzy search, I'm going to have to think of the best way of implementing that or seeing if there's an existing library to leverage.

alexwhitman:
Which bit of code is responsible for hiding the sidebar when it loses focus?  I think I've fixed all of the other focus issues but can't figure out this one.  Maybe I've been staring at the code for too long.

gnunn1:
It's this code in sidebar.d, not very intuitive but since it grabs the mouse any mouse click events outside of the sidebar trigger a close via passing a null to the sessionselected event.

```
    bool onButtonPress(Event event, Widget w) {
        trace("** Sidebar button press");
        //If button press happened outside of sidebar close it
        if (event.getWindow() !is null && lbSessions.getWindow() !is null) { 
            if (event.getWindow().getWindowStruct() != getWindow().getWindowStruct() && event.getWindow().getWindowStruct() != lbSessions.getWindow().getWindowStruct()) {
                notifySessionSelected(null);
            }
        }
        return false;
    }
```

alexwhitman:
Thanks, I've had a quick look at that at while I can get focus grabbed, the onButtonPress function doesn't get called.  I'll have another look later.

I think I've fixed a lot of the other things, it just needs to look better than it currently does but I'm no UI designer.

alexwhitman:
I think this is almost there.

Right now the search is only a simple string in string search but I'm planning on porting https://github.com/jeancroy/fuzzaldrin-plus/ to D and creating a package when I get some more time but for now the simple search works.

It still doesn't look great, could probably do with someone with a flair for design to take a look.  The images used at the moment are temporary.

@gnunn1 Feel free to give any feedback or suggestions.

alexwhitman:
I think this is as far as I'll be able to get it, at least for the next few weeks.  Known issues remaining are

- [ ] Needs visual go over.
- [ ] The ListBox should probably be in a ScrolledWindow but I've tried a few things I can't get the ScrolledWindow to have any height unless it's hard coded (which is then probably too big when there are only a few rows).  Hopefully fixable as part of visual updates.
- [ ] Doesn't have the fuzzy string search, just a case insensitive string in string search but it works for now and might be enough.  Can wait for feedback on this one.

gnunn1:
With regards to the ScrolledWindow, I usually just call setSizeRequest to set a reasonable default height for it. Here's an example from the password manager dialog I recently added:

```
        ScrolledWindow sw = new ScrolledWindow(tv);
        sw.setShadowType(ShadowType.ETCHED_IN);
        sw.setPolicy(PolicyType.NEVER, PolicyType.AUTOMATIC);
        sw.setHexpand(true);
        sw.setVexpand(true);
        sw.setSizeRequest(-1, 200);
```

I'll merge this during the weekend and do any cleanup needed for the UI.

alexwhitman:
I've added the list box into a ScrolledWindow with a default height which works.  I've tried to make it a bit cleaner by making the height depend on the height of the ListBox but calling `getHeight()` on the ListBox always returns `-1` even when the widget is visible.

PR #458, opened 2016-08-27
Correction with punctuation in the translation strings

igor-dyatlov:


gnunn1:
Please update localization in Weblate as per the the Readme

igor-dyatlov:
@gnunn1 With all due respect, I edited the file terminix.pot. Which serves as a template for all languages.
In all described the same commit **Add colon mark in the line translate** 
Localization will update according to the rules.

gnunn1:
Thanks, I missed the changes to the pot files. I'll freely admit I'm not a localization expert (@phw did all the great work here, I just right code), but the pot file is generated using extract-strings.sh from the source code so I believe changes to the pot file should be done in the relevant source files not in the pot file which is transitory. If I'm wrong about this let me know and we can discuss.

The changes to the russian translation should be done in weblate. While I can merge it into I typically then have to waste time to clean things up with conflicting merges between terminix and weblate repos and I'd really prefer not to be doing that. Hope that makes sense.

PR #466, opened 2016-09-07
Add option to not close window with last session

alexwhitman:
Inspired by an option in Firefox to not close the window when the last tab is closed, I've added the same functionality to terminix.  Quite often I'll hit `Ctrl+d` quickly to close terminals and end up closing terminix.

The default is to close the window so it acts in the same way it previously did, but turned off the option will open a new session in the window when the last session is removed.

The window can still be closed by pressing the close icon in the header bar.

### DIFF
--- a/data/gsettings/com.gexperts.Terminix.gschema.xml
+++ b/data/gsettings/com.gexperts.Terminix.gschema.xml
@@ -144,6 +144,11 @@
       <summary>Whether splitters use a narrow (default) or wide handle</summary>
       <description>If true, splitters between terminals use a wide handle. Only supported in GTK 2.16 and later.</description>
     </key>
+    <key name="close-with-last-session" type="b">
+      <default>true</default>
+      <summary>Whether the window should close when the last session is closed</summary>
+      <description>If false, when the last session is closed, a new session is opened.</description>
+    </key>
     <!-- Dark Theme -->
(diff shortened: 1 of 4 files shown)


gnunn1:
@alexwhitman Thanks, however note that the code in appwindow.d generally assumes there is always at least one session which admittedly was a poor decision on my part. You would need to add checks to all the spots where I'm just blindly taking what getCurrentSession gives me and using it. For example:

```
//Change name of session
        registerActionWithSettings(sessionActions, ACTION_PREFIX, ACTION_SESSION_NAME, gsShortcuts, delegate(GVariant, SimpleAction) {
            Session session = getCurrentSession();
            string name = session.name;
            if (showInputDialog(this, name, name, _("Change Session Name"), _("Enter a new name for the session"))) {
                if (name.length > 0) {
                    session.name = name;
                    updateTitle();
                }
            }
```

alexwhitman:
@gnunn1 I'm not really sure what you're referring to.  This change doesn't leave the app window open without any sessions, it opens a new session after the last one is closed.

With the option unset in the preferences:
- Open terminix
- Hit Ctrl+d in the terminal, terminal + session closes
- New session + new terminal opens

The code in your comment refers to changing the session name which I'm still able to do after going through the above steps.

gnunn1:
My misunderstanding.

alexwhitman:
Heh, no problem.  Thanks. :+1:

Issue #448, opened 2016-08-18
Scroll session by mouse wheel


AlexWayfer:
Hello!
Can you enable scrolling sessions by mouse-wheel scroll on "View session sidebar" button?

gnunn1:
So to be clear, you mean holding the mouse over the session button in the headerbar and using the mouse wheel to scroll through sessions? If I've gotten that right I'm not in favour of this, seems pointless versus just using keyboard shortcuts or am I missing something?

AlexWayfer:
> So to be clear, you mean holding the mouse over the session button in the headerbar and using the mouse wheel to scroll through sessions?

Yeah.

> seems pointless versus just using keyboard shortcuts or am I missing something?

Keyboard shortcuts are cool, but sometimes your hands only on mouse, and one scroll instead of click-move-click can be more convenient. For example, scroll on tab-bar in Chromium for tab-switch, or scroll on dash/desktop of Desktop Environments (GNOME, for example) for workspace-switch.

gnunn1:
Closed via PR from @alexwhitman which implements this.

PR #468, opened 2016-09-09
Change sessions by scrolling on sidebar button

alexwhitman:
This implements #448.  Simple enough but I get they you may not want to merge it based on what you said in the original issue.


gnunn1:
Thanks @alexwhitman, since you did the work and it's a trivial amount of work I have no issue with it.

PR #472, opened 2016-09-12
Add option to display the sidebar on the right

### LINKED ISSUE #470: Session switcher slide-out from the right
At the moment, the session switcher slides out from the left. It would be great if that was configurable. Having the switcher on the right, wouldn't cover terminal text which for the most part is always on the left.

Just a suggestion. Thanks.

alexwhitman:
Fixes #470


gnunn1:
Thanks, looks good to me.

PR #487, opened 2016-09-17
Make install script more robust

danielkza:
Using plain `cp` and `mkdir` will fail when using a non-standard umask
for either compiling Terminix or running the installation script. Fix that
by using the `install` command with explicit modes when necessary.

Additionally, ensure the desktop file cache and icon cache are properly
updated when either /usr or /usr/share is chosen as PREFIX.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -2,13 +2,14 @@
 
 if [ -z  "$1" ]; then
     export PREFIX=/usr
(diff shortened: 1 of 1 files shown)


gnunn1:
Thanks, I've been meaning to get to that for awhile!

PR #492, opened 2016-09-18
Fix quake mode focus

alexwhitman:
This seems to fix the quake mode focus issues, at least for me.  I've been running with this patch for a while and I've not seen any loss of focus.


alexwhitman:
Scratch that, it worked for a while and then stopped.

gnunn1:
If you are still having focus issues please open an issue for it and I'll have a look. On my system it appears to be working fine but wouldn't be surprised if different DE like unity require different fixes.

There is some code in experimental/source/x11.d that I cribbed from xfce that I can put back in to see if it helps. Once focus was working without it I took it out.

PR #498, opened 2016-09-20
Auto-hide the terminal title bar if its the only terminal in the session

shadowislord:
Adds a feature to hide the terminal's title bar automatically when its the only terminal in the session. Adds a check-box in the preferences to enable old functionality.


gnunn1:
Thanks, this is an interesting idea. I'll need to think about this a bit in terms of implementation, not sure if I like telling the terminal it's a single terminal versus setting an override to turn off the titlebar. This is somewhat a case of semantics but it's because I've tried to de-couple the terminals from knowing too much about what's going on in the session which the isSingleTerminal flag is kind of doing.

A couple of other points as follows:

a. IMHO, the preference should default to the existing behavior, i.e. don't turn off the titlebar by default. I'm also wondering if prefer the semantics of the preference be reversed, i.e. terminal-title-hide-when-single instead of terminal-title-show-when-single

b. One thing I think you missed is when the title is turned off, the context menu has additional items added to it. Look for the line "//Check if titlebar is turned off and add extra items" at line 1355 in your code in terminal.d.

Don't worry about changing your code until I've had some time to think about this a bit more. I'm not opposed to the idea, just need to think about the implementation for a day or two, I'm old and slow :)

gnunn1:
I thought about it and I'm OK with merging this change if you make the previously requested changes with the exception of the first one which can be ignored. Just leave it as terminal-title-show-when-single but make the default true to preserve the existing behavior.

shadowislord:
All the requested changes are now done. 
Let me know if you need anything else :+1:

PR #504, opened 2016-09-26
add dracular scheme

arshavindn:


gnunn1:
This theme is already available in the terminix theme repo here https://github.com/storm119/Terminix-Themes. In general, I'd prefer to keep the number of themes bundled with Terminix relatively small unless there is a special reason to include it.

at-hatv:
I have no idea about that repo, btw that's awesome, I've keep thinking that I have to convert other scheme type file to json to use with terminix

arshavindn:
I have no idea about that repo, btw that's awesome, I've keep thinking that I have to convert other scheme type file to json to use with terminix

PR #505, opened 2016-09-26
autotools: check for and link with libx11

dsboger:
I've used the pkg-config tooling to look for libx11 and added a corresponding linker flag. Works for me. More testing (especially cross-distro) appreciated.


gnunn1:
Thanks, I was trying to do it in configure.ac and have it flow into the makefile via DEP_LIBS and having no luck, now I see why.

PR #562, opened 2016-11-07
Might fix #561

### LINKED ISSUE #561: Double icons after update
After updating Terminix to new version I have faced this issue, there's a second fuzzy icon on the dock, even if I start it using the menu application, even if I start it using the shortcut.

image

bil-elmoussaoui:
More info here https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s05.html


gnunn1:
Thanks, however considering he said it was a new issue I don't see why this would fix it over the previous version?

bil-elmoussaoui:
Well, all the double icons issues are caused by a missing StartWMClass, which is used to detect the desktop file of the running application. There's a Gtk Window function to set a WMClass to the application window. Which might need to be added too

gnunn1:
I don't think native applications need this, if I run ```grep WM *.desktop``` on ```/usr/share/applications``` very few desktop files have the StartWMClass specified as far as I can tell.

PR #565, opened 2016-11-09
i18n: Obsolete strings were not automatically removed

phw:
Now extract-strings.sh updates the pot file so it will contain only existing strings.


phw:
This has also the advantage that the file location comments will stay up to date (previously they were accumulating all the locations a string has been at over time).

phw:
Just for reference, this broke with commit https://github.com/gnunn1/terminix/commit/9ef6685651be371b5a3115d57fab90344c4e7d49

PR #609, opened 2016-12-12
Mark a couple of strings for translation

dsboger:


gnunn1:
Thanks, why did you make the change for hbSide to set Hexpand to false from true? I had to set to true to fix an issue in Ubuntu with it's window control buttons on the left. If it's causing a problem for you, let me know and I'll try to find something that fixes things everywhere.

```hbSide.setHexpand(false);```

dsboger:
It's not a big problem. It's just a misalignment when resizing the preferences window (see attached image below). That's the simplest solution I could think of. Given that breaks other things, two other possible approaches are using a size group or even disabling resize for the prefwindow. What do you think?

image

phw:
@gnunn1 Once this gets merged can you update the POT file so we can update the translations on weblate before 1.4 release?

gnunn1:
Thanks @dsboger , I'll have a look at it tonite. 

@phw will do, the PR needs a few minor changes as well for the localization so that the HTML bold tags are excluded.

dsboger:
I fixed the HTML tag and removed the other commit.

gnunn1:
I'm making a lot of changes to this file and manually incorporated your fixes. I'll check it in shortly.

dsboger:
I see! Thanks!

PR #663, opened 2016-12-31
Fix build errors

f2404:
```
source/gx/terminix/terminal/exvte.d(90): Error: cannot implicitly convert expression (this.onNotificationReceivedListeners[this.onNotificationReceivedListeners.length - 1u].handlerId) of type ulong to uint
source/gx/terminix/terminal/exvte.d(141): Error: cannot implicitly convert expression (this.onTerminalScreenChangedListeners[this.onTerminalScreenChangedListeners.length - 1u].handlerId) of type ulong to uint
```

### DIFF
--- a/source/gx/terminix/terminal/exvte.d
+++ b/source/gx/terminix/terminal/exvte.d
@@ -77,7 +77,7 @@ public:
 	 *     summary = The summary
 	 *     bod = Extra optional text
 	 */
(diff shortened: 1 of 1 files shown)


gnunn1:
Is this a 32 bit issue, I just submitted a PR to GtkD that uses this type of event handling so I'll need to fix it there as well if it is.

f2404:
Yes, I encountered it trying to build terminix on a 32-bit system.

gnunn1:
Thanks!

PR #685, opened 2017-01-11
rpmlint warning for missing shebang

curantes:
This bash script is missing a shebang and is getting a warning from rpmlint report when build terminix as rpm package on OBS (http://build.opensuse.org).

### DIFF
--- a/data/scripts/terminix_int.sh
+++ b/data/scripts/terminix_int.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 # Adapted from vte.sh script, copyright notice from that ensues:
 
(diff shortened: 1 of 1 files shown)


gnunn1:
@curantes Just out of curiosity, is the script working for you? I don't get the impression many people have tried it and I suspect it could use some attention. For me, I've mostly switched to using triggers to parse out the command prompt since it avoids having to install anything on systems I SSH to.

curantes:
@gnunn1 I have not used/tested the script my self actually, I just wanted to make the rpmlint check happy during build. I could have just made a patch for it but thought it was better to add this here directly.

Thanks for accepting my Pull Request.

PR #689, opened 2017-01-13
Remove redundant style classes from sidebar

20115681:


gnunn1:
Originally I had those classes to fix some issues with the sidebar becoming transparent, however I see I'm setting the background color in the class terminix-session-sidebar so it should be OK.

PR #706, opened 2017-01-22
Add Nautilus shortcut, fixes #672

### LINKED ISSUE #672: Keyboard shortcut in Nautilus for Open in Terminix
Just an idea and not sure if it is possible, but it would be nice to have a keyboard shortcut inside Nautilus to trigger the "Open in Terminix" action for the currently shown folder.

I found this Nautilus plugin which does add a keyboard shortcut to Nautilus:

https://github.com/riclc/nautilus_backspace/blob/master/BackspaceBack.py

In theory it should be even possible to read the configured shortcut from the gsettings Terminix settings, which would make the shortcut easily customizable in Terminix itself.

bil-elmoussaoui:
Some code of the extension was still using Python2, modified that. Fixed some stuff and imported only needed libraries/functions.
The current shortcut is CTRL + Z
I opened this PR in order to see if it's possible to add a new configuration for that in Terminix itself, so I can change that to use the configured shortcut from DConf directly.  Or just if anyone has a better proposal 

Here's a small video that shows how it works
https://youtu.be/Sxi_8UggqGY

PS: works only with directories. FTP and SFTP are not supported (maybe it's possible) i will try to investigate more and push a fix in the future!

### DIFF
--- a/data/gsettings/com.gexperts.Terminix.gschema.xml
+++ b/data/gsettings/com.gexperts.Terminix.gschema.xml
@@ -1046,5 +1046,10 @@
       <default>'disabled'</default>
       <summary>Keyboard shortcut to cycle the terminal title styles</summary>
     </key>
+    <!-- Nautilus Shortcut for Open in Terminix -->
+    <key name="nautilus-open" type="s">
+      <default>'&lt;Ctrl&gt;z'</default>
+      <summary>Keyboard shortcut used in Nautilus for Open Here extension</summary>
+    </key>
   </schema>
 </schemalist>
--- a/data/gsettings/com.gexperts.Terminix.gschema.xml
+++ b/data/gsettings/com.gexperts.Terminix.gschema.xml
@@ -1048,7 +1048,7 @@
     </key>
     <!-- Nautilus Shortcut for Open in Terminix -->
     <key name="nautilus-open" type="s">
-      <default>'&lt;Ctrl&gt;z'</default>
+      <default>'&lt;Ctrl&gt;&lt;Shift&gt;t'</default>
       <summary>Keyboard shortcut used in Nautilus for Open Here extension</summary>
     </key>
   </schema>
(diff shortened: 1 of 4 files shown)


gnunn1:
Nice work! It's absolutely possible to add a shortcut for this to terminix itself and I'm happy to do it. One thing you will need to though is add an onChange to the GSettings object where you read the shortcut from do that if the shortcut changes you can update it.

If you add me as contributor to your Terminix repo I can add the shortcut change there so you can work on it and then submit everything as a single PR.

bil-elmoussaoui:
@gnunn1 You can already modify my commit, Github added a new option recently "Allow edits from maintainers" which allow the maintainers to modify the commits pushed to their repository 
Anyway I've added you as a collaborator ;)

gnunn1:
OK I've updated your master and patch-1 branches with the change needed to make the shortcut configurable in terminix. If you look at the Preferences->Shortcuts you will see a new section called Nautilus with one shortcut in it. 

The path in dconf for the shortcut is ```com/gexperts/Terminix/keybindings``` with the key ```nautilus-open```.

bil-elmoussaoui:
Perfect! I will compile my master branch and modify it ;) should be ready in few minutes!

bil-elmoussaoui:
@gnunn1 Ready to go! Everything work as expected :) the default shortcut is not perfect, but anyway everyone can customize it ;) maybe something like CTRL + SHIFT +T ? Not used by Nautilus :)

gnunn1:
@phw, any thoughts on the default shortcut to use for this since you opened the original issue?

phw:
Ctrl+shift+t seems to be a sane default. What about my comment above about reusing the existing _open_terminal code?

gnunn1:
@phw I'm probably missing something but I don't see the comment you are referring to?

phw:
@gnunn1: The comments in the changes, it is displayed above for me. Direct link: https://github.com/gnunn1/terminix/pull/706/files#r97214429

bil-elmoussaoui:
@phw sadly you can't reuse the same code as the Uri received on the get Uri is just a tuple that contains the filename, it's the reason why I used only one function for both (terminal shortcut and open file in the menuitem)

PR #716, opened 2017-01-26
Convert Find button to ToggleButton so it shows its state

### LINKED ISSUE #715: Use a Gtk.ToggleButton for Search
Search Button on the header bar is currently using Gtk.Button instead of Gtk.ToggleButton.
https://github.com/gnunn1/terminix/blob/master/source/gx/terminix/appwindow.d#L366

f2404:
Fixes #715

### DIFF
--- a/source/gx/terminix/appwindow.d
+++ b/source/gx/terminix/appwindow.d
@@ -102,7 +102,6 @@ import gx.terminix.preferences;
 import gx.terminix.session;
 import gx.terminix.sessionswitcher;
 import gx.terminix.sidebar;
-import gx.terminix.terminal.terminal : GxTerminal = Terminal;
 
 /**
  * The GTK Application Window for Terminix. It is responsible for
@@ -840,8 +839,7 @@ private:
             updateTitle();
             if (stateChange == SessionStateChange.TERMINAL_FOCUSED) {
                 Signals.handlerBlock(tbFind, _tbFindToggledId);
-                GxTerminal terminal = cast(GxTerminal)getActiveTerminal();
-                tbFind.setActive(terminal.isFindToggled());
+                tbFind.setActive(getActiveTerminal().isFindToggled());
                 Signals.handlerUnblock(tbFind, _tbFindToggledId);
             }
         }
--- a/source/gx/terminix/appwindow.d
+++ b/source/gx/terminix/appwindow.d
@@ -363,10 +363,11 @@ private:
         btnAddVertical.setFocusOnClick(false);
 
         // Add find button
-        Button btnFind = new Button("edit-find-symbolic", IconSize.MENU);
+        ToggleButton btnFind = new ToggleButton();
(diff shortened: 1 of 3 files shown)


gnunn1:
Thanks,  however there's a bit more involved. The find is per terminal so the toggle button has to change its state as the focus goes from terminal to terminal.

f2404:
Oh I see. Let me try to handle it.

f2404:
@gnunn1 Could you help me?
I see that there are focus in/out event handled by the Terminal class. On focus in, it could update the button state based on whether the Search pane is shown for the terminal. But it would need access to AppWindow's methods for that. Is it possible?

f2404:
Apparently, focus in event handler is not a proper place for that - switching between two terminals and opening find pane results in a sequence of focus in/out events and a segfault.

f2404:
Okay, I think I've the proper place - it's `onSessionStateChange()` in appwindow.d. I want to update Find button state on receiving `SessionStateChange.TERMINAL_FOCUSED` event, based on whether the focused terminal shows search pane or not. But, I think even calling `setProperty("active", false)` on the button triggers the 'toggled' handler.
Is there a way to temporarily disable the handler that was previously set by `addOnToggled()`?

gnunn1:
I made a pull request in GtkD that now returns the handlerId, you can use that to disable and re-enable it using ```Signals.handlerBlock``` and ```Signals.handlerUnblock```.

gnunn1:
Sorry I should be clearer, when you call one of the addOn... methods as of GtkD 3.5 or later, it returns the handlerId.

f2404:
@gnunn1 Thanks, that's a valuable addition!
In fact, http://api.gtkd.org/src/gtk/ToggleButton.html does say that `addOnToggled()` returns but I didn't notice that because the return value and it's meaning are not described there, unfortunately.

f2404:
With the 2nd commit it seems to be working properly for me.

gnunn1:
Thanks, one change request though. Instead of importing gx.terminix.terminal.terminal in appwindow.d, can you add the ```isFindToggled``` to the ITerminal interface in common.d. I'm trying to keep the Window from knowing too much about the Terminal by confining any necessary leakage to this interface.

f2404:
Thanks, this makes sense - please see the latest commit.

gnunn1:
Looks good, thanks for your work on this, much appreciated.

f2404:
Thanks for letting me do this. It's an interesting experience - programming in unknown language in a text editor with no IDE functionality ;)

gnunn1:
If you are interested in doing more, I highly recommend Visual Studio Code and the code-d plugin. It provides code completion as well as other features that make things much easier. If you want some help getting it up and running let me know and we can arrange a time to talk on IRC.

f2404:
Thanks for the tip, I will try it.

PR #734, opened 2017-01-31
Allow localizing manpages

### LINKED ISSUE #713: Man page localizations
Why no man page localizations?

phw:
Fixes #713 

This enables translation of manpages using po4a. [po4a](https://po4a.alioth.debian.org/) is required, however without po4a the install scripts will still work, it will just ignore manpage localization.

The process is as follows:

1. On changes to the manpage `extract-strings.sh` should be run. This was extended to also update the manpage translation template and translations in `data/man/po`
2. To translate into a new language just use `data/man/po/terminix.man.pot` to create a new PO file, name it `data/man/po/[locale].man.pot` (a German translation is included here as an example)
3. On installation `install.sh` will use the translations to place translated manpages in `share/man/[locale]/man1`

Once this is merged we should see to get this into Weblate also. I think we will have to ask Michal  from Weblate to add a new component for the manpage resources to the Terminix Weblate project.

### DIFF
--- a/data/man/po/de.man.po
+++ b/data/man/po/de.man.po
@@ -1,0 +1,375 @@
+# SOME DESCRIPTIVE TITLE
+# Copyright (C) YEAR Free Software Foundation, Inc.
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: \n"
+"POT-Creation-Date: 2017-01-31 23:37+0100\n"
+"PO-Revision-Date: 2017-01-31 23:23+0100\n"
+"Last-Translator: Philipp Wolfer <ph.wolfer@gmail.com>\n"
+"Language-Team: \n"
+"Language: de\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"X-Generator: Poedit 1.8.11\n"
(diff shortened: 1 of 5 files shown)


gnunn1:
I'll have to look at this one in more detail before accepting it and drop a note to the Weblate folks about how it would work.

One concern I have, this looks like it breaks up the man document into parapgraph sized chunks. Would this cause the man page to be localized as individual items and if so is that going to cause issues if the document needs to be re-structured for best translations, i.e. paras don't match on a 1:1 for basis?

phw:
> One concern I have, this looks like it breaks up the man document into parapgraph sized chunks.

All the serious translation tools I know perform some kind of segmentation for the text like this. One great advantage is that you can make better use of translation memory. Doing the German translation was an easy job, since I could source most of the translations from existing Terminix translations. The few parts that were actual work were the longer paragraphs. 

Also it actually improves the situation in cases where there are changes to the sources, since only parts that have changed need to be translated again. Just as in the source code gettext has no issue with moved parts or added paragraphs. If there is content added or changed in the source and the translation is not updated you will just end up with a partial translated document, whatever is untranslated will be kept in the source language.

> drop a note to the Weblate folks about how it would work.

In Weblate you can setup multiple components for a project, in this case a new component with a file mask of `data/man/po/*.man.po` and a base file of `data/man/po/terminix.man.po` should be added.

gnunn1:
@phw I've merged your changes, thanks very much for doing this. Do you mind e-mailing Michal at Weblate since you are probably more knowledgeable about what the ask is.

phw:
@gnunn1 Sorry for the late reply, but I have written Michal about it. I'll let you know once I hear back.

PR #751, opened 2017-02-08
Fix Quake mode activation

### LINKED ISSUE #749: Quake mode opening when new instance is set to focus window
I've been seeing an issue lately with Quake mode but I'm not sure when it started.  Under Ubuntu 16.04 I've configured a key combination to open Terminix in Quake mode.

If Terminix isn't running and I open a standard terminal window, triggering a quake mode window just focuses the terminal window.  If instead I first run Terminix in Quake mode and then open a standard terminal window, Quake mode works as intended.

It looks to be related to the "On new instance" setting as I have it set to "Focus Window".  If I set it to "New Window" then Quake mode opens as intended no matter what order Terminix was started in.

alexwhitman:
Fixes Quake mode activation when "On new instance" is set to "Focus Window" and a standard terminal window is opened before Quake mode is activated.

This seems to fix #749 for me but probably requires a further sanity check.

### DIFF
--- a/source/gx/terminix/application.d
+++ b/source/gx/terminix/application.d
@@ -399,7 +399,7 @@ private:
                 }
             } if (cp.preferences) {
                 presentPreferences();
-            } else {
+            } else if (!cp.quake) {
                 AppWindow aw = getActiveAppWindow();
                 if (aw !is null) {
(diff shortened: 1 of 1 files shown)


gnunn1:
@alexwhitman Thanks for this, I used your approach but just structured the code a bit different.

PR #759, opened 2017-02-14
Translator credits

phw:
Instead of a static string this uses the customary msgid “translator-credits” for the translator credts in the about dialog, as recommended at https://developer.gnome.org/gtk3/stable/GtkAboutDialog.html#gtk-about-dialog-set-translator-credits . This allows each translator to add himself to the credits, also makes the credits for the actual translation.

I have also updated the language files and added the existing contributors to the credits as they were credited in the PO files or commits already.


gnunn1:
Thanks @phw, much appreciated.

PR #761, opened 2017-02-14
hide searchentry by default

### LINKED ISSUE #756: Don't always show the searchentry on shortcuts settings
I think it's better to hide the search entry, and show it only when the user uses CTRL +F or add a Gtk.ToggleButton to the headerbar as it's done on the new design of Gnome settings :)

bil-elmoussaoui:
fixes #756 
On other Widgets
screenshot from 2017-02-14 14-56-06
screenshot from 2017-02-14 14-54-55
screenshot from 2017-02-14 14-55-37

@gnunn1 I would like to add a CTRL + F to reveal the search entry, but as it's my first attempt to write D code, some tips would be very helpfull :)

### DIFF
--- a/source/gx/terminix/prefeditor/prefdialog.d
+++ b/source/gx/terminix/prefeditor/prefdialog.d
@@ -50,6 +50,7 @@ import gtk.ListStore;
 import gtk.MenuButton;
 import gtk.MessageDialog;
 import gtk.Popover;
+import gtk.Revealer;
 import gtk.Scale;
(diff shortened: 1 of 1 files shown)


gnunn1:
Thanks, that's great. You can use addonKeyPress or addOnKeyRelease to the shortcut treeview to listen for Ctrl-F. Have a look at search.d for an example.

bil-elmoussaoui:
@gnunn1 Perfect! I will add that soon :)

PR #771, opened 2017-02-15
Fix bookmark type command icon

phw:


gnunn1:
Thanks, not sure I missed that one!

PR #804, opened 2017-02-27
Fixes #797

### LINKED ISSUE #797: open-terminix.py crashes Nautilus on startup
Traceback:

```
francisco@VoidCaster [12:24:58] [~] 
-> $ nautilus
sys:1: PyGIWarning: Nautilus was imported without specifying a version first. Use gi.require_version('Nautilus', '3.0') before import to ensure that the right version gets loaded.
Number of files is 0
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/gi/types.py", line 237, in mro
    return mro(cls)
  File "/usr/lib/python2.7/site-packages/gi/types.py", line 271, in mro
    for base in C.__bases__:
TypeError: get_background_items must return a sequence of Nautilus.MenuItem
**
ERROR:../../pygobject/gi/pygobject-object.c:993:pygobject_new_full: assertion failed: (tp != NULL)
[1]    3375 abort (core dumped)  nautilus
```

Using Nautilus 3.22.2 from ArchLinux. The crash only started happening after updating `terminix` to 1.5.0, so I suspect commit 0ef55af523762c1918753c98b1c2892ce6867478.
I will try to revert that commit and see if it helps.
Simply removing `/usr/share/nautilus-python/extensions/open-terminix.py` will avoid the crash (and remove the option "Open in Terminix" from nautilus.

bil-elmoussaoui:
This fix removes a condition that I've added in my latest commit to checking if Terminix is installed in the system or not.
Anyway, this fix it and I've also introduced a small condition to verify whether the GSettings are correctly found to avoid crashing Nautilus!

### DIFF
--- a/data/nautilus/open-terminix.py
+++ b/data/nautilus/open-terminix.py
@@ -19,7 +19,6 @@ require_version('Nautilus', '3.0')
 from gi.repository import GObject, Gdk, Gio, Gtk, Nautilus
 
 TERMINAL = "terminix"
-IS_INSTALLED = bool(call(['which', TERMINAL], stdout=PIPE, stderr=PIPE) == 0)
 
 
 def open_terminl_in_file(filename):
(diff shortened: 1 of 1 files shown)


gnunn1:
Thanks @bil-elmoussaoui!

PR #851, opened 2017-03-23
autotools: improve man page installation, generate localized man pages

### LINKED ISSUE #842: automake build does not process localized man pages
The automake build does not compile and install the localized manpages. It should do something similar to https://github.com/gnunn1/tilix/blob/master/install.sh#L113

I looked into it, but I don't have much experience with autotools and felt a bit lost. @dsboger could you take a look?

dsboger:
This adds a new build dependency for autotools builds (`po4a-translate`), required for localizing man pages.

EDIT: Works For Me, so more testing would be nice.
EDIT2: This should fix #842


gnunn1:
Thanks as always @dsboger

PR #903, opened 2017-04-27
Don't call query_info which might block

gicmo:
g_file_query_info is a sync API that will do IO and potentially
block if the network connection is slow (or dropped). If that
call blocks, it will freeze nautilus since the extension is called
from the main thread. Since we only use the resulting information
for the filename add a tooltip use the get_name() provided by
nautilus itself instead.


csoriano89:
Yeah, the point of nautilus providing the NautilusFile is that all the info is already cached so you don't have to query it from disk.

In any case, if necesary, _async can be done, but it can be problematic in these cases since it's used for adding items from the menu.

So basically +1 to the patch.

PR #929, opened 2017-05-16
Respect --title in Application title

MaskRay:


gnunn1:
I don't think it makes sense to use the overrideTitle there. The overrideTitile is passed in by updateTitle which getDisplayText then uses to process. The Window Title variable is meant to represent the terminal window title, not what is set in the chrome (i.e. Tilix), hence why it is fetched with ```vte.getWindowTitle()```.

Am I missing something or is there a specific problem you have noticed?

MaskRay:
Without this patch, the command line option `--title` does not set the application title.

I also wonder why both appwindow.d and terminal.d define `_overrideTitle` but the former does not seem to be changed anythere.

gnunn1:
Thanks, let me have a look at the ```--title``` option.

gnunn1:
I had a look at this and it seems to work as expected:

```tilix --title=Hello```

screenshot from 2017-05-17 18-29-50

MaskRay:
I mean the application title, not the tab title

gnunn1:
That switch sets the terminal title, not the application title. I don't have any plans to change it, however you can propagate the terminal title to the application title.

* Oops, no you can't because it doesn't use the overridden title. I'm going to open an issue for this and give it some thought.

gnunn1:
Added issue #930 for this, thanks for bringing it to my attention. Feel free to make any additional comments in that issue.

PR #951, opened 2017-06-01
autotools: Don't override DCFLAGS

jascrain:
The configure.ac sets DCFLAGS, which means that it's not possible to set custom DCFLAGS when building the program.  This changes configure.ac so that it will only set default DCFLAGS if it has not already been set externally.

Tilix in Debian was accidentally built with debugging enabled because of this.  See https://bugs.debian.org/863491.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -13,7 +13,7 @@ AC_ARG_VAR([DCFLAGS], [flags for dmd compiler])
 
 # Full optimization flags
 #DCFLAGS='-O -inline -release -version=StdLoggerDisableTrace'
-DCFLAGS='-O'
+: ${DCFLAGS='-O'}
 AC_SUBST([DCFLAGS])
 
 # Checks for programs.


dsboger:
I find it strange. I'm building tilix-git in Arch for some time now and I can easily override DCFLAGS in the make command line, like this:

`make DC='ldmd' DCFLAGS='-O -inline -release -version=StdLoggerDisableTrace'`

Does this not work for you?

jascrain:
Yes, overriding DCFLAGS with `make DCFLAGS=XXX` works.  But the Debian package builds more like this:

```
export DCFLAGS='-O -inline -release -g -version=StdLoggerDisableTrace'
./configure
make
```

This should also work.

gnunn1:
@dsboger Any objections to merging this, is it going to cause any issues elsewhere?

dsboger:
No objections at all! I was just curious. You see, I'm still learning this stuff.

gnunn1:
Thanks @dsboger, you may be learning but you are way ahead of me on autotools.

PR #966, opened 2017-06-07
Fix Nautilus decode unicode error

### LINKED ISSUE #965: Not working nautilus integration with unicode-named directories
Not working nautilus integration (Open with... feature) with unicode-named (if have in name at least one unicode or non-english symbol) directories. Arch Linux, Gnome 3.24.2, Nautilus 3.24.1, Tilix 1.6.1 - fresh compiled from AUR.

phw:
Fixes #965

See also http://python-gtk-3-tutorial.readthedocs.io/en/latest/unicode.html on details on pygtk encoding handling. The string object is guaranteed to use utf-8 encoding, so the decode is fine.


gnunn1:
Thanks @phw!

PR #972, opened 2017-06-16
Define the HOSTNAME Variable in tilix_int.sh

Belphemur:
On Debian the HOSTNAME variable wasn't define, using a call the the hostname bin populate it.


gnunn1:
I sit possible to add a check to only set the hostname variable if it is not already set?

Belphemur:
Good point, I've added a check to avoid any override.

gnunn1:
Thanks!

Issue #1018, opened 2017-07-16
Can't run tilix after building from source


Mike-Now:
After building from source I get an error:

```
2017-07-16T21:04:03.621:resource.d:findResource:53 Resource tilix/resources/tilix.gresource could not be found

(tilix:12548): GLib-GIO-ERROR **: Settings schema 'com.gexperts.Tilix.Keybindings' is not installed
```

It would be nice if the README file covered the necessary steps after building to start the application.

gnunn1:
Looks like you missed the step to run install.sh as per the README which says this after building:

```
The application depends on various resources to function correctly, run sudo ./install.sh to build and copy all of the resources to the correct locations. Note this has only been tested on Arch Linux, use with caution.
```

shinenelson:
I ran into the same issue even after running `install.sh`. Though I was smart enough to read the code of the script to find that I could pass an argument to set `PREFIX` to a custom directory than `/usr` (that could be mentioned in the `README` too).

I tried 3 different options on 3 isolated bare VMs : 
1. With the default option for `PREFIX` installing to `/usr` (and with `sudo` privileges)
2. Installing to `$HOME/.local` (`PREFIX=$HOME/.local`) (without `sudo` privileges)
3. Installing to custom build directory (`PREFIX=build`) (without `sudo` privileges; within the same directory as the cloned repository) and then adding the `build` directory to my `PATH`.
None of the above options worked for me.

I read through the `install.sh` script and the script looks to be installing the application correctly. I verified that the above mentioned file that is not found is indeed there in the exact same location that the binary says is not found.

I looked at the `data/resources/tilix.gresource.xml` file that generates `data/resources/tilix.gresource`. It has no mention of `com.gexperts.Tilix.Keybindings`. Could it be that the source code isn't compiling the Keybindings properly?

Or maybe the binary is looking for a different location than the location the file is in?

What could be the actual reason? I'm not familiar with D to dig into the source code to find the bug here. Otherwise, I'd be happy to help in any way I could.

gnunn1:
The keybindings are not managed by the resource file but by the schema file which is installed separately. What was the output of install.sh, were any errors shown?

A common problem is not having the tool to compile the resources file installed.

Also for running tilix I've only tested it with installing it to ```\usr```. Not sure if it would being installed elsewhere, however if you get it working I'd be more then happy to take a PR for it.

shinenelson:
[...]

PR #1020, opened 2017-07-16
Fix issue with running Tilix after manual compiling to custom directory

shinenelson:
This fixes Issue #1018.

Changes : 
* Locked the installation of gschema to the `/usr` directory as there are no way around it to install gschema to a custom location.
* Added checks for `PREFIX` and `sudo` privileges before installing gschema
* Added some warnings for users that try to install to custom directories without `sudo` privileges.
* Modified user id checks for `root` from string to numeric.
* Updated README with relevant information about installing to custom directories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ dub build --build=release
 ```
 
 The application depends on various resources to function correctly, run `sudo ./install.sh` to build and copy all of the resources to the correct locations. Note this has only been tested on Arch Linux, use with caution.
+Note : `install.sh` will install Tilix to your `/usr` directory. If you are interested in installing Tilix to a custom location, you can specify the `PREFIX` as an argument to `install.sh` script (e.g : `./install.sh $HOME/.local` will install Tilix into `$HOME/.local`). While installing to custom directories seem to be working fine ([with one exception](https://github.com/gnunn1/tilix/wiki/Installing-to-custom-directory)) and doesn't require `sudo` privileges; please be warned that you will have to re-compile the glib schema manually and this would require `sudo` privileges.
 
 Note there is also experimental support for autotools, please see the wiki page on [autotools](https://github.com/gnunn1/tilix/wiki/Building-with-Autotools) for more information.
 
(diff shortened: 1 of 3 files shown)


gnunn1:
Thanks, I'll have a look in the morning. I think one potential issue is that I use the install script to create the archive package in which case the schema can't be locked to use. See the data/pkg/archive folder.

shinenelson:
I don't think the lock-in should affect you as long as you don't provide `sudo` privileges.

In fact, you could remove the extra step for removing the compiled schema from the `data/package/archive/createReleaseArchive.sh` script since the schema won't be compiled at all with this change.

shinenelson:
According to https://github.com/gnunn1/tilix/issues/1018#issuecomment-315646510, we wouldn't need this fix anymore. All that needs to be done is to append the custom `$PREFIX/share` directory to `$XDG_DATA_DIRS`.

shinenelson:
Just a heads up here. If you're installing tilix to a custom directory and appending `$PREFIX/share` to `$XDG_DATA_DIRS` as mentioned in https://github.com/gnunn1/tilix/pull/1020#issuecomment-315649501, you'd need to make sure that `XDG_DATA_DIRS` are initialized during Xsession initialization. Otherwise, Tilix wouldn't run out of the Applications menu. It would work fine when invoked from a terminal (because you're probably going to be appending `$PREFIX/share` to `$XDG_DATA_DIRS` during your shell intialization and not Xsession initialization), but it wouldn't work from the Applications Menu.

I spent the last 3 days trying to figure the problem out ( what sense did `kernel: traps: tilix[<pid>] trap int3 ip:xxxxxxxxxxxx sp:xxxxxxxxxxxx error:0` really make? ).
The issue is because `dbus` initializes the `XDG_DATA_DIRS` environment variable during Xsession initialization.
I got the heads up (idea) from a freedesktop mailing list [post](https://lists.freedesktop.org/archives/dbus/2016-July/016960.html) with a similar issue related to `dbus`.

You could approach this in 2 different ways (both require `sudo` privileges anyway) : 
1. You provide `sudo` privileges during install; and let `glib-compile-schema` install the `tilix.gresource` install to the main list of `gschema`.
2. You could be a hacker and add `$PREFIX/share` to your Xsession initialization in `/etc/X11/Xsession.d/60x11-common_xdg_path`. (Any file in `/etc/X11/Xsession.d/` would seem to be fine, but this file seems to be the most appropriate as it does exactly that - initialize the `XDG_DATA_DIRS` environment variable)

PR #1072, opened 2017-08-14
[RFC] add total number of sessions in application title

syedelec:
I use this patch to show the total session number in the application title. A usecase is having the current session / total session, for example 1/2 :

screen


gnunn1:
Thanks, can you change ```VARIABLE_TOTAL_SESSION_NUMBER``` to ```VARIABLE_SESSION_COUNT``` and the localization to ```Session Count```

syedelec:
Updated variables and localization

PR #1086, opened 2017-08-21
removed unneeded terminix entries

sogalboogalsky:
Hi,
Just a small PR in which I have removed old remaining "Terminix" strings.
I haven't touch to Flatpack files as some links probably need to be updated before.


gnunn1:
Thanks!

Issue #582, opened 2016-11-23
Use flatpak's HostCommand to avoid sandboxing


gnunn1:
When terminix is run within a flatpak, it is sandboxed and does not have the necessary permissions to run as a full terminal application. Christian Hegert was kind enough to point out how the embedded terminal in gnome-builder deals with it by using the flatpak HostCommand DBus message.

This issue is to do the same in terminix so that it can be packaged as a flatpak.

Links that Christian provided:

Pty setup: https://git.gnome.org/browse/gnome-builder/tree/plugins/terminal/gb-terminal-view.c#n238

HostCommand(): https://git.gnome.org/browse/gnome-builder/tree/libide/subprocess/ide-breakout-subprocess.c#n1448

gnunn1:
Been doing some work on this but turning into a pain in the ***, shelving it until I'm more in the mood or have an epiphany.

gnunn1:
This has been implemented thanks to @jaje

PR #1125, opened 2017-09-17
Improve Flatpak build

jaje:
Tilix now can work from flatpak sandbox.

- Shell runs outside sandbox (in host environment)
- Uses PATH from host
- Sometime it crashes


experimental/flatpak/com.gexperts.Tilix.json
```diff
@@ -0,0 +1,65 @@
...
+                                "sed -i 's/\\/usr\\/bin\\/tilix/tilix/g' /app/share/dbus-1/services/com.gexperts.Tilix.service"
+                        ],
+                        "build-options": {
+                                "build-args": ["--share=network"]
+                        },
+                        "sources": [
+                                {
+                                        "type": "git",
+                                        "url": "../../.git",
+                                        "branch": "master"
```
jaje: This presumes that this pull is merged into master, if you want to try it change branch and/or url.

jaje:
Issue: #582

gnunn1:
Not sure how can I test this, looks like flatpak requires xattr for the --user switch now, ugh. Not keen on compiling a kernel with it enabled nor running builds as root.

jaje:
The --user switch has nothing to do with wherever it runs as root or not (it never does), it only switch form shared (/var/lib/flatpak) to user (~/.local/share/flatpak) repo.
Btw I thing xattr support depends on filesystem, not kernel.

gnunn1:
Thanks, adding user_xattr to fstab does indeed work. Now when I run build.sh the install.sh script fails as it cannot find xdg-desktop-menu. Is there a dependency missing?

```
...
Running: ./install.sh /app
Your system is missing command xdg-desktop-menu, please install xdg-utils
Error: module tilix: Child process exited with code 1
```

jaje:
No, this happens when you don't change branch as I said in comment. If you use the build script without this change it will build Tilix from master branch, not from this pull. 
There are possible 2 solutions: 
1) clone it locally and merge into master
2) change "branch" to "flatpak" (in experimental/flatpak/com.gexperts.Tilix.json)

gnunn1:
Thanks it is working now, however I'm having some issues with it running without crashing. Notably bringing up preferences for example causes it to quit. Having said that, this is a huge improvement over what I have now so I'm to go ahead and merge it.

Thanks for making the effort on this, I really appreciate it as I was stuck on this for awhile.

PR #1126, opened 2017-09-18
Fix Flatpak build

jaje:
- ~~Do not crash when opening window (it still sometimes crash when closing window)~~
- Keep stdio usable
- Use env vars from host, so that more commands work (for example `flatpak remote-list`)
Issue #582

### DIFF
--- a/source/gx/tilix/appwindow.d
+++ b/source/gx/tilix/appwindow.d
@@ -1765,17 +1765,6 @@ public:
         }
         //Create an initial session using default session name and profile
         createSession(gsSettings.getString(SETTINGS_SESSION_NAME_KEY), prfMgr.getDefaultProfile());
-
-        static if (FLATPAK) {
-            // Inside flatpak sandbox first session complains:
-            //      sh: cannot set terminal process group (-1): Inappropriate ioctl for device
-            //      sh: no job control in this shell
-            Session init_session = getSession(0);
-            if (init_session !is null) {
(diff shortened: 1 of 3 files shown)


jaje:
- Command which need pty (such as sudo) should now work correctly
- Randomly crashes

jaje:
I thing that we had same problem as described [here](https://github.com/GNOME/gnome-builder/blob/a41d6e8d9ac360c03848992b54eec27107b73d29/libide/subprocess/ide-breakout-subprocess.c#L1226), should be fixed by https://github.com/gnunn1/tilix/pull/1126/commits/cc3cb9e998e0bfe2ff53d91916bfbf166b915980

jaje:
OK, I think that most crashes are fixed by https://github.com/gnunn1/tilix/commit/cc3cb9e998e0bfe2ff53d91916bfbf166b915980

gnunn1:
Just double checking, is this ready to be merged?

jaje:
Yes, it is ready.
There are some todos, but since I have no clue how to workaround https://github.com/gnunn1/tilix/pull/1126/commits/0242f72c424394bf31c15cea6a4f0485e34ee6ac, I can't fix them.

jaje:
Sorry, forgot about O_CLOEXEC (so that sudo as 1st command worked, but not as 2nd ect.), but now it is really ready, I won't push more commit into this.

gnunn1:
Thanks, I'll do a quick test Tonite and merge it in. Once again, thanks for your work on this, I really appreciate it.

gnunn1:
Thanks again for your work on this, my apologies for the delay.

Issue #937, opened 2017-05-24
Feature Request:  support for viewing current/running process in title


s4n-gt:
ConEmu on Windows and iterm2 on MacOS have a feature to show running process.
(ref: https://www.reddit.com/r/linux4noobs/comments/6cu425/show_current_process_in_terminal/)

it would be nice to have this functionality in tilix.

gnunn1:
What do they do for processes with excessively long names?

s4n-gt:
[sorry for late replay]

for long names, conEmu only shows first 63 characters.

![conemu](https://cloud.githubusercontent.com/assets/28922018/26506159/57047456-426d-11e7-83c1-c72a487952ee.png)

gnunn1:
I played around with this, the primary obstacle is being notified when a child process is created. Unless the window title changes or some other activity happens where VTE triggers an event tilix doesn't know that a process kicked off. For example, when running nano no events are triggered in tilix.

I'd prefer not to resort to polling the shell process periodically as that just chews up cycles. If someone more familiar with kernel capabilities knows of a way to get notified of new processes that doesn't require root access feel free to let me know. I didn't see any obvious candidates in a brief search.

s4n-gt:
I am not familiar with kernel.

please keep this alive for someone who is knowledgeable in this stuff.

gnunn1:
I think I'm going to have to resort to polling for this, I believe that's what iterm2 is doing.

gnunn1:
I have added an initial implementation of this feature, however since it is using threads to do the polling I have disabled it by default for now until I get a chance to test it before. I have only briefly tested it but it seems to work quite well in this limited testing. Note that the polling period is set to 250ms at the moment and I don't think it's feasible to set it lower.

I still have some things to optimize as well. At the moment all terminals receive notifications even if it is for a process that is not there. Need to figure a way to bind signals selectively in D that is clean.

If someone wants to try it, you need to enable the ```USE_PROCESS_MONITOR``` flag in constants.d and then re-compile tilix.

gnunn1:
Also, when enabled you will notice a new variable called ```Process``` in the list to be used for titles.

gnunn1:
Sorry for the spam, just reminding myself of a few things. I wonder if I could use this feature to drive notifications on systems that don't have the fedora VTE notification patch?

Also, might be an idea to add trigger support for it as well.

gnunn1:
I've enabled this feature by default, let me know if there are any issues.

[...]

PR #1134, opened 2017-09-27
additional fix for #937

s4n-gt:
sorry for claiming that it didn't work. I didn't understand that application title and terminal title are two different things as I have always hidden terminal titlebar.

P.S: I have noticed another segmentation fault after adding a new variable to the terminal title if at least one variable exists. I don't know to trace it.

### DIFF
--- a/source/gx/tilix/customtitle.d
+++ b/source/gx/tilix/customtitle.d
@@ -165,7 +165,6 @@ private:
             case ViewMode.EDITOR:
                 setHexpand(true);
                 setVisibleChildName(PAGE_EDIT);
-                eTitle.grabFocus();
         }
     }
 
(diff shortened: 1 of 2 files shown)


s4n-gt:
https://github.com/gnunn1/tilix/issues/937#issuecomment-331688566

gnunn1:
Sorry for the delay, I will try merging this tomorrow night.

gnunn1:
I'm not sure about the appTitle thing, I assume you are doing that because the application title has a variable the terminal title does not so therefore the title change never gets emitted to the window?

I'm thinking it might just be better to either improve the emit to detect changes to the individual values or simply emit it all the time. I don't want the terminal to be responsible for checking if other titles need to be changed.

Let me think about it for a bit.

s4n-gt:
yes, If the terminal title has some not changing variable (`id`, `column`) and application title has some frequently changing variable (`title`, `process`), then application title won't change (unless you switch between terminal, session etc.).

gnunn1:
I've merged your changes and made a couple of small tweaks afterwards. Thanks for your help with this.

PR #1159, opened 2017-10-20
additional support for active process #937

s4n-gt:
~It is not efficient but I am comfortable with it.~

~Not sure, how to handle `MonitorEventType`.~
~https://github.com/s4n-gt/tilix/blob/f91c70f5f3e7bae2fddea899e49f19d400ee5eb1/source/gx/tilix/terminal/monitor.d#L140-L146~

please let me know if it needs more comments on the code.

(It is my second D PR, so, I am happy to accept any negative or positive comment.)
(I won't mind if it takes some time before you have enough time to review it)

### DIFF
--- a/source/gx/tilix/terminal/activeprocess.d
+++ b/source/gx/tilix/terminal/activeprocess.d
@@ -57,6 +57,7 @@ class Process {
         if (!Process.pidExists(pid)) {
             return false;
         }
+        // Need updated version.
         string[] tempStat = parseStatFile();
         long pgrp = to!long(tempStat[3]);
         long tty = to!long(tempStat[5]);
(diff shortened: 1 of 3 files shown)


s4n-gt:
please wait. Maybe, I have found a better way :)

s4n-gt:
@ gnunn1
done. I have tried with 40 open terminals and it usage very low amount of CPU (upto 0.50%). also there could be some optimization e.g when it is in the idle state.

please consider that I am still a beginner and I have no D experience. so, you will need some clean up.

gnunn1:
I don't see any problems with your code, I'll merge it if it is good to go and then tweak if necessary after the merge. Is that OK?

s4n-gt:
I have no more changes, so, please do it :)

Issue #617, opened 2016-12-15
Few issues with Quake mode


bil-elmoussaoui:
Hello,
I just tried Quake mode on Arch, using the latest git version of terminix and on xorg  and gnome 3.22
When i set both width and height to max using the preferences dialog, the dimensions changes correctly but when i resize it, nothing happens :( Is that intended ?

Also, using the max width, there's a pixel on left and on right that are not covered by the terminal widget
![capture d ecran de 2016-12-15 03-46-27](https://cloud.githubusercontent.com/assets/7660997/21210151/1517ca8a-c279-11e6-8193-2d236f84c1d6.png)

One last issue, if i do set the height & width to max of quake window, close terminix and open terminix again on Quake mode; i can't change the height, and the height of the window override the gnome shell bar (i don't know what we call it exactly). i think that this issue is somehow related to the first one. 
![capture d ecran de 2016-12-15 03-50-25](https://cloud.githubusercontent.com/assets/7660997/21210232/a0166fc4-c279-11e6-9419-a11fcee2052d.png)

Thank you very much!

gnunn1:
Thanks for the report, I'll have a look tomorrow night.

gnunn1:
Just a note to myself, I wonder if I should be using getMonitorWorkarea instead of getMonitorGeometry.

gnunn1:
The last issue, when you set the width and height to 100%, is because the window manager decides to switch the window to full-screen. I've tried a few things like using the monitor workarea coordinates instead of the geometry or preventing it by returning true when the window state changes to fullscreen but no luck.

Instead I'm going with the easy solution, only allowing the height to be set to 90% of the available height instead of maximum.

gnunn1:
As for the 1 pixel border on the left and right, I can confirm this but also can confirm I'm setting the right coordinates. Again, I think this is a Window Manager thing and not something I can resolve.

I also tested xfce4-terminal with it's drop-down mode and it exhibits the same symptoms.

bil-elmoussaoui:
Thank you! :+1:

f2404:
@gnunn1 I think both of these issues are specific to a certain WM; they are not happening on Cinnamon or xfwm4. Maybe we should submit bugs against Gnome WM?

gnunn1:
[...]

PR #1164, opened 2017-10-24
Allowed terminal height in quake mode to be max 100%, up from 90%

will-nolan:


gnunn1:
100% height causes problems with window managers hence why it is not supported. See this comment and surrounding issue:

https://github.com/gnunn1/tilix/issues/617#issuecomment-267458402

If you fix that I'm happy to re-consider.

PR #1202, opened 2017-11-27
Fix crashes caused by x11 gdk extern declarations

pgkos:
This PR fixes the extern declarations of x11 functions from the gdk library.

The function symbols in `__gshared extern(C)` were compiled as global function pointers initialized to zero, which caused the relocation table in `libgtk-3.so.0` to contain invalid entries for these particular functions.

This is fixed by prepending a prefix to avoid a name clash and introducing an alias (just like in the gtkd library).


gnunn1:
Thanks for the PR, much appreciated.

PR #1241, opened 2018-01-06
Consolidate VTE version checks

alexwhitman:
Moves all of the version checks to constants for consistency and makes it easier to patch to compile on Ubuntu 17.10.


gnunn1:
Thanks @alexwhitman, that's exactly where I was going with these changes.

Issue #1252, opened 2018-01-10
Not working with Python3 in Arch


zgwerby:
Using python-nautilus (pointing to python3 and not python2-nautilus, it's different packages) the menu item does not appear.

Running python open-tilix.py from /usr/share/nautilus-python/extensions/ I get the message

`
Traceback (most recent call last):
  File "open-tilix.py", line 8, in <module>
    from urllib import unquote
ImportError: cannot import name 'unquote'
`

gnunn1:
At the moment it's not written for Python 3, please use ```python2-nautilus``` as per the dependencies for Arch Linux.

bil-elmoussaoui:
@gnunn1 I can update the code to work with both python versions it shouldn't be that complicated

gnunn1:
Thanks @bil-elmoussaoui, I've re-opened the issue.

gnunn1:
@zgwerby Please try the updated version and let me know if it works for you

zgwerby:
@gnunn1 Yes, I've been paste new script - working fine. But I have new problem, like in #1250 issue. And I think it global bug - extensions like nautilus-admin (after fixing by me on my machine, like near pull request), that use python-nautilus bindings have duplicate menu entry too. Sorry for my poor language.

PR #1253, opened 2018-01-10
Fix import error on python3

bil-elmoussaoui:
Fixes the nautilus-python issue. Tried it on my machine and works correctly 

Fixes https://github.com/gnunn1/tilix/issues/1252

### DIFF
--- a/data/nautilus/open-tilix.py
+++ b/data/nautilus/open-tilix.py
@@ -5,13 +5,13 @@
 
 from gettext import gettext, textdomain
 from subprocess import PIPE, call
-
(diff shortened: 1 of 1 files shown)


gnunn1:
Is this compatible with both pythin 2 and 3?

bil-elmoussaoui:
Yep, the whole code was using the python 2.7 which is somehow near to the python3 code style. 
The issue was caused by few internal libraries that were renamed in python3

gnunn1:
Thanks, I've merged the code

PR #1259, opened 2018-01-17
Normalize text strings in ProfileEditor

f2404:
Some text strings in ProfileEditor stand out by using CamelCase notation. Propose to make them look like other strings that have only one capital letter.


gnunn1:
Thanks!

PR #1317, opened 2018-03-03
Fix transform func misstyping '/' -> '\'

p1scescom:
This fixs paste trasnform function.
Its motto is replaceing CRLF & CR to LF but it replace '/r/n' & '/r' -> '/n'.
So Fix ' / ' -> ' \ '.


gnunn1:
Thanks for catching that and taking the time for a PR

PR #1351, opened 2018-03-25
Add auto-oriented splitting (Request for feedback from community)

henvo:
Hello,

I came across this awesome project a couple of weeks ago and have been using this terminal emulator since then. So first of all thank you for developing and maintaining this great project!

Before using Tilix in GNOME I was using i3wm and I really liked the "auto-tiling" feature that allowed me to split screens vertically and horizontally with one key binding (mod + Enter). So I implemented it also for my new favorite terminal emulator.

This PR adds a new key to split a terminal depending on the current height and width.

Here's the result:
tilix-pr
Here I pressed the "auto-tiling" key three times.

I would be glad to get some feedback on this PR and maybe see this feature implemented in the next
version of Tilix!

### DIFF
--- a/data/gsettings/com.gexperts.Tilix.gschema.xml
+++ b/data/gsettings/com.gexperts.Tilix.gschema.xml
@@ -1013,6 +1013,10 @@
       <default>'&lt;Ctrl&gt;&lt;Alt&gt;d'</default>
       <summary>Keyboard shortcut to add new terminal down</summary>
     </key>
+    <key name="session-add-auto" type="s">
+      <default>'&lt;Ctrl&gt;&lt;Alt&gt;a'</default>
+      <summary>Keyboard shortcut to add new terminal automatically</summary>
+    </key>
     <key name="win-view-sidebar" type="s">
       <default>'F12'</default>
       <summary>Keyboard shortcut to view session sidebar</summary>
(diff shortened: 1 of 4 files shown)


gnunn1:
Thanks for the PR, I'm not sure I would use it but I do not have any strong objections to merging it. I'll leave this open for a few days to let others have opportunity for feedback.

develop7:
I'd love to use it!

phw:
I'm also for merging, the change is small enough. I definitely would give this a try if it is useful for me, as this is the logic I apply most of the times when splitting a terminal.

data/gsettings/com.gexperts.Tilix.gschema.xml
```diff
@@ -1013,6 +1013,10 @@
       <default>'&lt;Ctrl&gt;&lt;Alt&gt;d'</default>
       <summary>Keyboard shortcut to add new terminal down</summary>
     </key>
+    <key name="session-add-auto" type="s">
+      <default>'&lt;Ctrl&gt;&lt;Alt&gt;x'</default>
```
phw: What about `Ctrl+Alt+a` as default (a for just "add")?

henvo: Yes, fine by me. Should I change this in my pull request?

gnunn1: Yes please

gnunn1:
Thanks for the contribution @henvo, much appreciated.

PR #1353, opened 2018-03-26
Update man page for --window-style and --display options

f2404:


f2404:
It seems Travis is broken, it fails on
`Fetching gtk-d 3.7.5 (getting selected version)...`

gnunn1:
Thanks!

gnunn1:
I had to revert this as it is causing problems with localization, please ensure that extract-strings and install.sh can be run after making this change. When I look at I believe the problem is that the strings are not getting updated correctly in the tilix file.

f2404:
I think I found a problem with extract-strings - please see #1359

PR #1359, opened 2018-04-03
Fix manpage translations update

f2404:
`po4a-updatepo` does not accept multiple output files provided with a single `-p` option.
Update the script to call `po4a-updatepo` per each translation file.

### DIFF
--- a/extract-strings.sh
+++ b/extract-strings.sh
@@ -60,8 +60,13 @@ do
 done
 
 # Update manpage translations
+echo "Updating manpage translations..."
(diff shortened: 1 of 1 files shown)


f2404:
Hi @gnunn1, do you see any problems with this PR? It's been almost a week...

gnunn1:
The only problem is a lack of time, sorry. I'll have a look at it later today.

gnunn1:
It's been updated, thanks for your patience and taking the time to submit these PRs.

f2404:
Thank you!

PR #1423, opened 2018-05-21
Language rework

comradekingu:


data/man/tilix
```diff
@@ -75,13 +75,13 @@ Do not show client side decorations and toolbar.
 Do not show window borders.
 .TP
 .B \-\-new\-process
-Start an additional Tilix instance as a new process. By default and as per GTK+ 3 guidelines, Tilix maintains a single process that all windows belong to. This option forces Tilix to start as a separate process. This is not recommended and is intended for debugging purposes only, using this option will prevent different Tilix windows from communicating with each other. 
+Start an additional Tilix instance as a new process. By default and as per GTK3+ guidelines, Tilix maintains a single process that all windows belong to. This option forces Tilix to start as a separate process. This is not recommended and only intended for debugging purposes. Its use will prevent different Tilix windows from communicating with each other.
```
dsboger: In the [official GTK+ blog](https://blog.gtk.org/), they always use GTK+ [version_number], so maybe it should keep that format?

data/man/tilix
```diff
@@ -75,13 +75,13 @@ Do not show client side decorations and toolbar.
...
 .B \-\-new\-process
-Start an additional Tilix instance as a new process. By default and as per GTK+ 3 guidelines, Tilix maintains a single process that all windows belong to. This option forces Tilix to start as a separate process. This is not recommended and is intended for debugging purposes only, using this option will prevent different Tilix windows from communicating with each other. 
+Start an additional Tilix instance as a new process. By default and as per GTK3+ guidelines, Tilix maintains a single process that all windows belong to. This option forces Tilix to start as a separate process. This is not recommended and only intended for debugging purposes. Its use will prevent different Tilix windows from communicating with each other. 
 .TP
 .B \-\-geometry=GEOMETRY
 Set the window size; for example: 80x24, or 80x24+200+200 (COLSxROWS+X+Y).
 .TP
 .B \-q \-\-quake
-Open a window in quake mode or toggle existing quake mode window visibility.
+Open a window in quake mode or toggle existing Quake mode window visibility.
```
dsboger: You missed a quake :).

data/man/tilix
f2404: I agree. It isn't GTK3+, it's GTK+ version 3.

data/man/tilix
```diff
@@ -21,28 +21,28 @@ Show the version of Tilix as well as the versions of dependent components.
...
 .B \-p \-\-profile=PROFILE_NAME
-Set the starting profile using the name of one of the existing profiles. If a name is passed that does not correspond to an existing profile, the default profile will be used.
+Set the starting profile using the name of one of the existing profiles. If it matches none of those, the default profile will be used.
 .TP
 .B \-t \-\-title=TITLE
 Set the title of the new terminal.
 .TP
 .B \-s \-\-session=SESSION_NAME
-Open the specified session file which has been previously saved. Session files are saved from within Tilix, they store the complete layout of the session enabling the user to re-load it when required. Note that only the layout and layout options are saved, this will not restore the terminal to where it was left off, rather new terminals are created.  
+Specify a previously saved session to open. They store the complete layout of the session, saved and re-loaded from within Tilix when required. Note that only the layout and layout options are saved, this will not restore the terminal to where it was left off, rather new terminals are created.
```
f2404: Who are "they"?

gnunn1:
I'm closing this PR as the OP hasn't responded to the comments.

comradekingu:
@gnunn1 Didn't see that, if you open it again I will update it.

data/man/tilix
comradekingu: How about:
"Specify a previously saved session to open and re-load when required. Only the layout and layout options are saved"


"this will not restore the terminal to where it was left off, rather new terminals are created."
This part i don't understand.

f2404: > Specify a previously saved session to open and re-load when required.

"when required" is not needed - once you've specified a previously saved session, it will be opened and reloaded.

> This part i don't understand.

This part means that the terminals themselves with their states, such as clients running or scrollback history, are not saved, in fact. Only the layout is saved, and based on that new terminals are created.

comradekingu: OK, how about:
"Specify a previously saved session to re-load the layout and layout options from. This will not restore the terminal to where it was left off, rather new terminals are created."

f2404: This sounds right.

However, I'm not sure what was the reason to change the existing text in the first place, but I guess I'll let @gnunn1 decide on that.

PR #1430, opened 2018-05-31
[WIP] Add initial support for snap

guilhem:
Add support for packaging application in snap.
This should help to provide the best experience to user (lib VTE etc).

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,76 @@
+name: tilix
+version: git
+summary: A tiling terminal emulator for Linux using GTK+ 3
+description: |
+  Tilix is a tiling terminal emulator which uses the VTE GTK+ 3 widget with the following features:
+    - Layout terminals in any fashion by splitting them horizontally or vertically
+    - Terminals can be re-arranged using drag and drop both within and between windows
+    - Terminals can be detached into a new window via drag and drop
+    - Input can be synchronized between terminals so commands typed in one terminal are replicated to the others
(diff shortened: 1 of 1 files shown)


guilhem:
This doesn't work for now with error:
```
object.Exception@../../../../.dub/packages/gtk-d-3.8.2/generated/gtkd/gtkd/Loader.d(125): Library load failed (libvte-2.91.so.0): /snap/core/current/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.25' not found (required by /usr/lib/x86_64-linux-gnu/libvte-2.91.so.0)
```

FFY00:
Why are you using DMD as the compiler? Snap packages are intended to contain release binaries. DMD is the reference compiler, it uses Digital Mars' backend which it's not the best in terms of optimization. The optimal choice would be GDC but LDC would also be acceptable.

guilhem:
@FFY00 really good question, II don't really know this ecosystem and only try to guess by some docs on web :)
but none of the others compliers performed to build tilix...
Help really needed ^^

Vild:
This is how tilix is built on ArchLinux: https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/tilix
But I would recommend LDC over GDC as GDC is still at 2.068.2.

FFY00:
@gnunn1 can we close this? It has been over a year without any updates. This could be reopened in the future if someone is interested.

gnunn1:
Closing since no work has progressed, if someone ones to pick it up feel free. Thanks @FFY00

PR #1456, opened 2018-07-06
Only show VTE config warning dialog when opening preferences 

ximion:
Hi!
On first run, Tilix shows a warning dialog on VTE configuration issues on a lot of distributions ( https://gnunn1.github.io/tilix-web/manual/vteconfig/ )

Since Tilix is the default terminal emulator on a couple of
distributions (PureOS, Budgie, ...), throwing a warning message into the user's face is not a
great initial experience. This also applies to cases where Tilix is not the default, of course.
So, those distributions would disable the warning message, which is also
not great, because it hides an issue. Changing the VTE configuration distribution-wide is also not an option most of the time.

This PR implements a compromise of these positions: Tilix will not show the warning message when it is run, but instead when the user opens the preferences dialog.
This should be reasonable, because users wondering about the drawbacks of the VTE (mis)configuration like the folder name not being remembered when splitting terminals will go to Tilix' preferences first and then see the issue.
Most other user who just want to use a terminal right now, won't see the warning.

This should prevent us at PureOS in particular from disabling the message completely.
Thank you for considering!

### DIFF
--- a/source/gx/tilix/application.d
+++ b/source/gx/tilix/application.d
@@ -124,6 +124,7 @@ private:
     ImageSurface isFullBGImage;
 
     bool warnedVTEConfigIssue = false;
+    bool shouldWarnVTEConfigIssue = false;
 
     bool useTabs = false;
 
@@ -787,6 +788,9 @@ public:
         });
         preferenceDialog.showAll();
         preferenceDialog.present;
+
+        // show VTE config warning dialog if configuration issues were detected
+        showVTEConfigIssueWarning(preferenceDialog);
     }
 
     void presentProfilePreferences(ProfileInfo profile) {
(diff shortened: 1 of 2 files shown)


ximion:
@phw It's not broken behavior on the distribution's side, just a different configuration option. Actually reading `/etc/profile.d/` for non-login shells is the "wrong" behavior, and I wonder why Fedora changed it.
You can find a great explanation on this behavior in https://askubuntu.com/questions/247738/why-is-etc-profile-not-invoked-for-non-login-shells (it says Ubuntu, but applies to Debian just as much).

In that light I was actually going to disable the warning in Tilix, because the default in Debian (or Arch) won't change (not quickly for sure, but also highly likely never, because it would break user assumptions).
This patch is a compromise, so Tilix users are still made aware that they can improve the terminal's behavior if they want to do so.

ximion:
@gnunn1 Okay, in that case we will override the notification for now. Maybe there is a solution for this that does not only work on very few distributions, I will take a look at this at a later point in time.

gnunn1:
Why not source it in ```/etc/bash.bashrc``` and the equivalent for zsh, I think it's ```/etc/zsh/zshrc```? Is that problematic in terms of wanting to depend on the upstream Debian packages in order to avoid custom packages?

ximion:
@gnunn1 Editing textfiles via package installations is very error-prone, so the only way to do that would be to convince the bash and zsh Debian maintainers to source files from a specific directory, or to add code to load the vte.sh file in particular.
If they are open to that solution, this might work.

phw:
Relevant review of PureOS, which mentions this issue: https://opensource-usability.blogspot.com/2018/07/first-impressions-of-pureos.html?m=1

gnunn1:
The other option is that I could add a hidden option to make the "use login shell" the default for profiles. Having said that, I'll freely admit I don't fully understand the pros/cons of login versus non-login shells beyond profile.d gets sourced.

ximion:
FTR, this only work on Debian and derivatives for GNOME Terminal because we carry a patch using older upstream behavior since forever: https://salsa.debian.org/gnome-team/gnome-terminal/blob/debian/master/debian/patches/Provide-fallback-for-reading-current-directory-if-OS.patch

Issue #957, opened 2017-06-04
Nautilus Plugin issue/improvements


bilelmoussaoui:
For now , the nautilus plugin will crash both Nautilus & Tilix (if you run Nautilus from terminal) if you try to open tilix using the keyboard shortcut on a google-drive folder (the ones added in gnome 3.20)

The reason why, is that we assume that the file scheme is always gonna be a "file" which is not always the case.
The plugin already handle that part of code for the menu items; I've already a working code that handle both cases, but i didn't create a PR yet until i'm sure that it works as expected with ftp & sftp, and as i don't have any access to any of those; some testing would be great :)

Here's the current code i've 
```python
# -*- coding: UTF-8 -*-
# This example is contributed by Martin Enlund
# Example modified for Tilix
# Shortcuts Provider was inspired by captain nemo extension

from gettext import gettext as _, textdomain
from subprocess import call
from urllib import unquote
from urlparse import urlparse

from gi import require_version
require_version('Gtk', '3.0')
require_version('Nautilus', '3.0')
from gi.repository import Gio, GObject, Gtk, Nautilus


TERMINAL = "tilix"
TILIX_KEYBINDINGS = "com.gexperts.Tilix.Keybindings"
GSETTINGS_OPEN_TERMINAL = "nautilus-open"
REMOTE_URI_SCHEME = ['ftp', 'sftp']
textdomain("tilix")


def open_terminal(uri_data):
    username = uri_data.username
    hostname = uri_data.hostname
    port = uri_data.port
    path = uri_data.path
    scheme = uri_data.scheme
    if scheme in REMOTE_URI_SCHEME:
        if username:
            cmd = 'ssh -t {0}@{1}'.format(username, hostname)
        else:
            cmd = 'ssh -t {0}'.format(hostname)

        if port:
            cmd += " -p {0}".format(port)

        if path:
            cmd += ' cd "{0}" ; $SHELL'.format(path)

        call('{0} -e "{1}" &'.format(TERMINAL, cmd), shell=True)
    elif scheme == "file":
        if path:
            call('{0} -w "{1}" &'.format(TERMINAL, path), shell=True)
        else:
            call("{0} &".format(TERMINAL), shell=True)


class OpenTilixShortcutProvider(GObject.GObject,
                                Nautilus.LocationWidgetProvider):

    def __init__(self):
        source = Gio.SettingsSchemaSource.get_default()
        if source.lookup(TILIX_KEYBINDINGS, True):
[...]

PR #1467, opened 2018-07-10
Fixes #1458

### LINKED ISSUE #1458: "Open Remote Tilix Here" fails if path contains space
When in an SFTP location in Nautilus, the context menu provides an option called "Open Remote Tilix Here". Usually, it works and does exactly what it says. However, if the path on the remote side contains a space, the shell is opened in the default location on the remote (usually the user's home directory) and this error message is printed to the terminal:

    bash: line 0: cd: /path/to/location/on/remote/containing/a%20space: No such file or directory

Seems like a string that's not been converted back from URL format.

bilelmoussaoui:
The issue was due to 
```python
value = '{0} cd "{1}" ; $SHELL'.format(value, result.path)
```
The value of `result.path` is now passed to `unquote` function before. And the spaces also have a \ before so it works correctly on the terminal.
Tried that on my personal FTP server and everything works correctly.

### DIFF
--- a/data/nautilus/open-tilix.py
+++ b/data/nautilus/open-tilix.py
@@ -37,7 +37,7 @@ def open_terminal(uri):
             value = "{0} -p {1}".format(value, file_obj.port)
         if Nautilus.FileInfo.create_for_uri(uri).is_directory():
             _dir = unquote(file_obj.path).replace(" ", "\ ")
-            value = '{0} cd "{1}" ; $SHELL'.format(value, _dir)
+            value = '{0} "cd {1}" ; $SHELL'.format(value, _dir)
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
This PR also adds a simple verification to avoid having an old version of Tilix with the latest version of the nautilus extension crash if the keybinding gsettings doesn't exist.

bilelmoussaoui:
The latest commits fixes #957 too. This way the Nautilus extension will work on all the cases. The extension works with FTP/SFTP/File schemes for both cases: the extension menu and the keyboard shortcut.

bilelmoussaoui:
Ready to be merged.

gnunn1:
Looking at the original issue it looks like the user is still having problems, should this be merged now or wait a bit?

PR #1507, opened 2018-08-28
Port to meson

bilelmoussaoui:
This a work in progress in order to allow other to play around with #1502 

What doesn't work:
- Still doesn't build correctly, see #1502 
- The translations still need some work
- The man page needs to be installed (need some work in order to translate the man file too). Gettext doesn't support man page yet :/

Before merging:
- Update the manual installation instructions on the website and the README file.
- Remove the install.sh & uninstall.sh, which should be handled now by meson.


data/meson.build
```diff
@@ -0,0 +1,107 @@
...
+    output:  meson.project_name() + '.desktop',
+    input: 'pkg/desktop/' + meson.project_name() + '.desktop.in',
+    po_dir: join_paths(meson.source_root(), 'po'),
+    type: 'desktop',
+    install: true,
+    install_dir: 'share/applications'
+)
+
+# Validate AppData file
+appstreamcli = find_program('appstream-util', required:false)
```
ximion: Any reason for using `appstream-util` over the reference `appstreamcli` here?
If you prefer to use asutil, I would like to know the reason for that. In any case, please don't call it appstreamcli and also use its `validate-relax` method in the build process, otherwise compiling Tilix will be a pain if asutil is installed and more up-to-date than what Tilix was last tested with.

ximion: @bilelmoussaoui That's not true, it supports `content_rating` since AppStream version 0.11.0 which was released in May 2017, more than a year ago.

bilelmoussaoui: Well, we had that issue with FeedReader using the latest Fedora

ximion: Can you point me at the issue? The only thing I'm aware of that was recently changed was empty content_attribute tags being allowed instead of setting every attribute to none explicitly.

bilelmoussaoui: That's exactly what I was talking about

ximion: Please report these things next time! Also keep in mind that an empty `content_rating` is not yet supported by all tools, so using it at the moment is still a bit risky (I actually only noticed this because of an error related to it).

bilelmoussaoui: Reported that! Thanks for your work on appstream! :)

gnunn1:
@bilelmoussaoui Can you restore the autotools stuff in this PR as I would like to have meson and autotools in parallel for a bit. I will then go ahead with merging it.

gnunn1:
I've done an initial import of these changes to the repo, unfortunately since there the addition of the new files plus the deletion of autotools was merged together I couldn't cherry pick commits to a merge and ended up just adding the files manually but crediting @bilelmoussaoui in the commit message.

I'm closing this as since it's been merged.

bilelmoussaoui:
Sorry I wasn't able to edit my PR sooner. Totally forgot about it!! Thanks for doing that :)

FFY00:
@gnunn1 next time you can put him directly in the commit :smile:

https://help.github.com/articles/creating-a-commit-with-multiple-authors/

PR #1537, opened 2018-09-30
Don't hardcode datadir

bilelmoussaoui:


gnunn1:
Dies ```datadir``` need to be set in the main meson build somewhere as I'm not seeing where it gets set in this PR?

bilelmoussaoui:
It's set by default to /share. It can also be set using --datadir argument during the build configuration

gnunn1:
Thanks, merged. Appreciate all you help with this meson stuff as my expertise is much more on the Java side (i.e. Maven) when it comes to build systems.

PR #1538, opened 2018-09-30
configure dbus service file

bilelmoussaoui:
This is the latest missing piece to make the meson build system switch working perfectly!
The tests stuff could be handled later

### DIFF
--- a/data/dbus/com.gexperts.Tilix.service
+++ b/data/dbus/com.gexperts.Tilix.service
@@ -1,3 +1,0 @@
-[D-BUS Service]
-Name=com.gexperts.Tilix
-Exec=/usr/bin/tilix --gapplication-service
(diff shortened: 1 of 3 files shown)


bilelmoussaoui:
I think we should backport this to autotools unless you're willing to remove it completely soon. The build won't work at all due to the missing file (added a .in suffix)

gnunn1:
You are welcome to but can't we have the file as it is now for autotools and the .in version for meson? If you want to backport it though I don't have any issue with that.

bilelmoussaoui:
my autotools skills are close to 0 :p but yeah we can do that! :) as it will be removed anyway in the near future!

bilelmoussaoui:
Done :)

gnunn1:
Thanks!

PR #1553, opened 2018-10-14
Redesign the icon for GNOME 3.32

bugaevc:
GNOME 3.32 is going to ship with a new icon style — see http://jimmac.musichall.cz/blog/2018-07-18-detail-considered-harmful/ and https://gitlab.gnome.org/GNOME/Initiatives/issues/2. Tilix is not a core GNOME app, but it'd be nice if it also got updated to make it fit well with the new design.

I'm not a designer (at all), but I use Tilix a lot so I thought I'd step forward and try to design a new icon according to the guidelines & style. Here it is:
image

For the reference, here's how the new GNOME icons look (from the GitLab link above):

image

And here's what it looks like in the [Icon Preview app](https://gitlab.gnome.org/ZanderBrown/icon-tool):

image

Along with the actual style change, they're also abandoning individual icon sizes (48x48 and so on) and instead going with a single scalable icon, so I've done the same here. I could have broken something, so please check it works.


gnunn1:
Thanks, this looks great. I'll merge it in the next few days after trying it out.

tristan957:
@bugaevc really nice job on the redesign

rugk:
Cross-post:
> oh, that Maps icon really looks way too much like Google Maps. Seriously, it looks like exactly the same colors!

https://social.wiuwiu.de/web/statuses/100924286227325517

PR #1574, opened 2018-11-06
Minor Fix: Rename Tilix context menu entries to match GNOME Terminal

cookiengineer:
This is just a minor fix and the only thing it will do is basically rename the context menu entries in a way that they look the same as the ones for GNOME Terminal.

The introduced else condition for the remote scheme behaves the same like before, but will show "Open Remote Tilix" and "Open Local Tilix" like it does for GNOME Terminal on GVFS mounts.

PS: Not sure if this is a too opinionated pull request, but I hope this gets merged ^_^

### DIFF
--- a/data/nautilus/open-tilix.py
+++ b/data/nautilus/open-tilix.py
@@ -111,14 +111,14 @@ class OpenTilixExtension(GObject.GObject, Nautilus.MenuProvider):
                 uri = file_.get_uri().decode('utf-8')
                 item = Nautilus.MenuItem(name='NautilusPython::open_remote_item',
                                          label=_(u'Open Remote Tilix'),
-                                         tip=_(u'Open Remote Tilix In {}').format(uri))
+                                         tip=_(u'Open Remote Tilix in {}').format(uri))
                 item.connect('activate', self._menu_activate_cb, file_)
                 items.append(item)
 
             filename = file_.get_name().decode('utf-8')
             item = Nautilus.MenuItem(name='NautilusPython::open_file_item',
-                                     label=_(u'Open In Tilix'),
-                                     tip=_(u'Open Tilix In {}').format(filename))
+                                     label=_(u'Open in Local Tilix'),
+                                     tip=_(u'Open Tilix in {}').format(filename))
             item.connect('activate', self._menu_activate_cb, file_)
             items.append(item)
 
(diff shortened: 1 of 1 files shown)


gnunn1:
@bilelmoussaoui you are more familiar with the plugin then I am, if you don't mind can you have a quick look before I merge this to see if you have any concerns?

data/nautilus/open-tilix.py
```diff
@@ -128,14 +128,23 @@ def get_background_items(self, window, file_):
...
+                                     tip=_(u'Open Remote Tilix in this directory'))
             item.connect('activate', self._menu_activate_cb, file_)
             items.append(item)
 
-        item = Nautilus.MenuItem(name='NautilusPython::open_bg_file_item',
-                                 label=_(u'Open Tilix Here'),
-                                 tip=_(u'Open Tilix In This Directory'))
-        item.connect('activate', self._menu_background_activate_cb, file_)
-        items.append(item)
+            item = Nautilus.MenuItem(name='NautilusPython::open_bg_file_item',
```
bilelmoussaoui: I don't see why this is needed here. This is only shown for ftp/sftp mounted folders. There's no "local" thing here at all.

bilelmoussaoui:
@gnunn1 Just took a look. There's a new menu item for the remote URI scheme. Not sure why it's needed and what it does do exactly.

gnunn1:
@cookiengineer Any response to @bilelmoussaoui comment about the new menu item?

cookiengineer:
@gnunn1 I probably misunderstood the API in an incorrect manner.

I'm not sure on what to do now in order to fix it. The idea was that it behaves the same as the standard GNOME Terminal, so that when you are in a gvfs-mounted folder, that the `remote` variant will connect via SSH session to the endpoint, and that the `local` variant will open Tilix in the local mount folder.

@bilelmoussaoui As I was implementing it incorrectly, how would I achieve something like this correctly?

bilelmoussaoui:
@cookiengineer Let me try that tomorrow and get back with what should be done and if it's ready to be merged.

gnunn1:
Can I close this or are you still looking at this @bilelmoussaoui ?

gnunn1:
Closing since this isn't going anywhere.

PR #1587, opened 2018-11-29
Path matching for auto profile switching

CromFr:
Currently the rule `:~/myfolder` will match both `~/myfolder/` and `~/myfolderwithextrachars/`

This solution compares folder by folder if the path matches


gnunn1:
Thanks!

PR #1600, opened 2018-12-19
UnsafePasteDialog: Make sure pasted command fits on the prompt

bilelmoussaoui:
See https://twitter.com/BabyWogue/status/1075527836612612096

Before the patch, the app crashes on Wayland/X11
```
(tilix:7412): Gdk-CRITICAL **: 00:22:11.876: gdkdisplay-wayland.c:1402: Unable to create Cairo image surface: invalid value (typically too big) for the size of the input (surface, pattern, etc.)
fish: “tilix” terminated by signal SIGSEGV (Address boundary error)
```

Results after the patch
screenshot from 2018-12-20 00-31-10

Doesn't happen anymore on Wayland.


gnunn1:
Thanks!

PR #1608, opened 2018-12-27
Improve the appdata file

bilelmoussaoui:
Per the specs
https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html

### DIFF
--- a/data/appdata/com.gexperts.Tilix.appdata.xml.in
+++ b/data/appdata/com.gexperts.Tilix.appdata.xml.in
@@ -32,7 +32,7 @@
   </description>
 
   <screenshots>
-​   <screenshot type="default">https://gnunn1.github.io/tilix-web/assets/images/gallery/tilix-screenshot-1.png</screenshot>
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
It would be nice to add update_contact too. But I don't have any idea which email to use there. Plus, the release tags should be kept up to date for future releases. They can be translated, they are shown on GNOME Software, Discover (KDE), Flathub and so on.

gnunn1:
The version number is proliferating into too many spots, I need to figure out a way to get it down to updating in one spot before merging this as otherwise I'm likely to miss spots when creating releases.

bilelmoussaoui:
You can configure those files to update the version number. The appdata file has to be updated manually as you need to set the release date & changelog.

bilelmoussaoui:
@gnunn1 what are those files you had in mind? I can fix them using Meson ;)

gnunn1:
Don't worry about it, it's just constants.d and this appdata file which is manageable particularly at my current pace of releases. I was conflating the GtkD versioning in there as well which is a separate issue.

PR #1630, opened 2019-01-11
gtk: stop setting inApp menu

### LINKED ISSUE #1629: Tilix AppMenu isn't dropped
GNOME 3.32 dropped appMenus, but this is what happens on Tilix 1.8.9

screenshot from 2019-01-11 01-47-12

bilelmoussaoui:
Per the GNOME Design initiative to remove the usage of the in-app menu and replace it with a primary menu in the application itself. We already have everything moved to the primary menu.
Fixes #1629


phw:
Shouldn't the entire installAppMenu be removed?

bilelmoussaoui:
A part of the function is used to register the actions. Not sure if it's done somewhere else too. Indeed the whole app menu stuff should be removed.
Will take a closer look at the code asap

gnunn1:
Thanks for taking care of that, tested and it works great.

bilelmoussaoui:
@gnunn1 Can we get a minor release for this? The AppMenu is not used anymore on the beta GNOME (Fedora Rawhide).

PR #1633, opened 2019-01-18
Meson: validate appdata without network access

bilelmoussaoui:
Currently, we validate the appata file using appstream-util which requires network access to fetch the screenshots and validate them. 
A lot of distros out there run tests without network access, this will make the appdata test fail and they might patch Tilix downstream.


gnunn1:
Thanks!

PR #1638, opened 2019-01-24
rename tabs (sessions) with double click

benpicco:
This adds the possibility to rename tabs when double clicking them, just as Terminator does it.

screenshot at 2019-01-24 02-04-19


gnunn1:
Thanks for this, there were a couple of minor issues which I fixed up. Not a complaint just an FYI. I changed the following:

- Switched to using a Stack to toggle between label and editor
- Focused editor when it becomes active
- Added the titleeditor which is used elsewhere
- Added support for escape and return keys

Give it a test if you have time and let me know if everything looks good.

PR #1650, opened 2019-02-12
App icon: update to most recent GNOME guidelines

bertob:
The most recent icon guidelines recommend bottom-aligning wide shapes.

This PR fixes the alignment and also makes the icon a bit more distinct from GNOME Terminal by adding a second split, and a more unique symbolic.


gnunn1:
Thanks, my apologies for the delay!

PR #1655, opened 2019-02-25
Meson: clean the build system

bilelmoussaoui:
- Don't use dub to retrieve dependencies anymore, it's correctly detected now
- Add LINGUAS file as it's needed by i18n module
- Bump meson version and use some features from it like '/' instead of join_paths.


bilelmoussaoui:
LDC fails to link a lot of things here :/ Not sure if anyone can reproduce that

data/meson.build
```diff
@@ -2,32 +2,42 @@
 # install the icons
 install_data(
     'icons/hicolor/scalable/apps/@0@.svg'.format(meson.project_name()),
-    install_dir: join_paths(get_option('datadir'), 'icons/hicolor/scalable/apps')
+    install_dir: get_option('datadir') / 'icons' / 'hicolor' / 'scalable' / 'apps'
 )
 install_data(
     'icons/hicolor/scalable/apps/@0@-symbolic.svg'.format(meson.project_name()),
-    install_dir: join_paths(get_option('datadir'), 'icons/hicolor/scalable/apps')
+    install_dir: get_option('datadir') / 'icons' / 'hicolor' / 'symbolic' / 'apps'
```
BrainBlasted: Instead of `get_option('datadir') everywhere, why not declare `datadir =  get_option('datadir')` and then `pkgdatadir = join_paths(datadir, 'dir_name')` for common directories like datadir/tilix?

data/meson.build
```diff
@@ -2,32 +2,42 @@
...
+)
 
 # Validate desktop file
 desktop_file_validate = find_program('desktop-file-validate', required:false)
 if desktop_file_validate.found()
   test (
     'Validate desktop file',
     desktop_file_validate,
-    args: join_paths(meson.current_build_dir (), '@0@.desktop'.format(meson.project_name()))
+    args: desktop_file.full_path()
```
BrainBlasted: If you need to use `full_path()` here, you should probably add `depends: desktop_file`. `full_path()` apparently loses target tracking information, so meson _could_ try to run this test before the file actually exists. It's not likely, but I'd be careful because I've run into it in other scenarios.

BrainBlasted: You might also be able to pass `desktop_file` as args, though I'm not sure if `merge_file()` returns the right type of object for that.

data/meson.build
bilelmoussaoui: Thanks for the review Chris! Will fix that soon

worldofpeace:
ping @bilelmoussaoui 

###### I could pick this up for you if you don't have the time for it.

bilelmoussaoui:
@worldofpeace please do that.  I'm not sure when I can update this one. Thanks!

bilelmoussaoui:
Closing this in favor of @worldofpeace's PR.

PR #1669, opened 2019-03-17
Add Yaru scheme

madsrh:
Add Yaru (Ubuntu) color scheme

See issue here: https://github.com/ubuntu/yaru/issues/1270

image


gnunn1:
I don't normally take new themes in the main repo but given the general popularity of Ubuntu I think this makes sense to include. Thanks for taking the time to put it together.

Issue #1619, opened 2019-01-04
Meson build doesn't install man page


bilelmoussaoui:
Currently, the meson build doesn't install the man page. Also, gettext doesn't support translating man pages for now :/ 

This need to have a proper fix before ditching autotools completely.

Issue #1693, opened 2019-04-19
Fail to build with meson on openSUSE Tumbleweed


curantes:
I try to build Tilix with meson but during configure phase I get below error:
`data/meson.build:99:0: ERROR:  Output file name must not contain a subdirectory.` 

I am not sure what the issue is. Any help would be great.

part of build log with configure
```
Build started at 2019-04-19T20:18:20.004975
Main binary: /usr/bin/python3
Python system: Linux
The Meson build system
Version: 0.49.2
Source dir: /home/abuild/rpmbuild/BUILD/tilix-1.9.0
Build dir: /home/abuild/rpmbuild/BUILD/tilix-1.9.0/build
Build type: native build
Project name: com.gexperts.Tilix
Project version: 1.8.5
Native D compiler: ldc2 (llvm 1.13.0 "LDC - the LLVM D compiler (1.13.0):")
Build machine cpu family: x86_64
Build machine cpu: x86_64
Message: Using compiler llvm
Found pkg-config: /usr/bin/pkg-config (0.29.2)
Determining dependency 'gtkd-3' with pkg-config executable '/usr/bin/pkg-config'
Called `/usr/bin/pkg-config --modversion gtkd-3` -> 0
3.8.5
Called `/usr/bin/pkg-config --cflags gtkd-3` -> 0
-I/usr/include/d/gtkd-3/
Called `/usr/bin/pkg-config gtkd-3 --libs` -> 0
-L-L/usr/lib64/ -L-lgtkd-3 -L-ldl
Called `/usr/bin/pkg-config gtkd-3 --libs` -> 0
-L-L/usr/lib64/ -L-lgtkd-3 -L-ldl
Dependency gtkd-3 found: YES 3.8.5
Determining dependency 'vted-3' with pkg-config executable '/usr/bin/pkg-config'
Called `/usr/bin/pkg-config --modversion vted-3` -> 0
3.8.5
Called `/usr/bin/pkg-config --cflags vted-3` -> 0
-I/usr/include/d/gtkd-3/
Called `/usr/bin/pkg-config vted-3 --libs` -> 0
-L-lvted-3 -L-L/usr/lib64/ -L-lgtkd-3 -L-ldl
Called `/usr/bin/pkg-config vted-3 --libs` -> 0
-L-lvted-3 -L-L/usr/lib64/ -L-lgtkd-3 -L-ldl
Dependency vted-3 found: YES 3.8.5
[...]

PR #1703, opened 2019-05-11
[🐛 FIX] meson subdir placement & install-man-pages

andrei-pavel:
Fixes:
#1619 Meson build doesn't install man page
#1693 Fail to build with meson on openSUSE Tumbleweed

The way you install bits and pieces with meson is through external scripts. Usually python, but why not reuse what's in `install.sh`?! That's why `install-man-pages.sh` is so convenient. `data/scripts` is the directory to place scripts which should be installed with tilix, but where else to put it? Placing it in root directory crowds it too much.

The changes in the portion where `PREFIX` is set is so that PREFIX can be set solely through environment variables before running a script.


data/scripts/install-man-pages.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/env sh
+
+# Set PREFIX.
+if [ -z  "${1+x}" ]; then
+    if ! [ -z "${PREFIX+x}" ]; then
```
FFY00: We should use `-n` instead of `! -z` for the sake of simplification and readability.

data/scripts/install-man-pages.sh
```diff
@@ -0,0 +1,26 @@
...
+        PREFIX=/usr
+    fi
+else
+    PREFIX=${1}
```
FFY00: `${1}` should be quoted, otherwise the script will fail if the argument includes a space.

data/scripts/install-man-pages.sh
```diff
@@ -0,0 +1,26 @@
...
+fi
+export PREFIX
+
+echo "Installing man pages"
+install -d ${PREFIX}/share/man/man1
```
FFY00: There's no need to invoke `install` to create both the directory and the file. This can be accomplished by simply using the `-D` flag.

data/scripts/install-man-pages.sh
```diff
@@ -0,0 +1,26 @@
...
+install -m 644 data/man/tilix ${PREFIX}/share/man/man1/tilix.1
```
FFY00: Change `-m` to `-Dm`, see the previous comment.
`${PREFIX}` should be quoted, see comment above.

data/scripts/install-man-pages.sh
```diff
@@ -0,0 +1,26 @@
...
+gzip -f ${PREFIX}/share/man/man1/tilix.1
+
+if type po4a-translate >/dev/null 2>&1; then
+    for f in data/man/po/*.man.po
+    do
```
FFY00: To keep the script style we should move the `do` to the line above.

install.sh
```diff
@@ -108,20 +112,7 @@ install -d ${PREFIX}/share/dbus-1/services
...
-if type po4a-translate >/dev/null 2>&1; then
-    for f in data/man/po/*.man.po
-    do
-        LOCALE=$(basename "$f" .man.po)
-        install -d ${PREFIX}/share/man/${LOCALE}/man1
-        po4a-translate -k 0 -f man -m data/man/tilix -p data/man/po/${LOCALE}.man.po -l ${PREFIX}/share/man/${LOCALE}/man1/tilix.1
-        gzip -f ${PREFIX}/share/man/${LOCALE}/man1/tilix.1
-    done
-fi
+. "$(dirname $(realpath ${0}))/data/scripts/install-man-pages.sh"
```
FFY00: The quotes should be around `${0}` instead of `$()`. `$()` already prevents word splitting, it doesn't need quotes. `${0}` can include spaces so it does need to be quoted.

data/scripts/install-man-pages.sh
andrei-pavel: Commendable motive. And I definitely don't mean to rock the boat. However, just want to point out that they're not perfect synonyms. When using the `[ ]` or the `test` builtins to evaluate expressions, in `sh`, `bash` and `zsh`, `[ -n ]` / `test -n`, that is `-n` with an empty string beside it so in this case with `${PREFIX}` empty, would evaluate to true, which usually is the opposite of what you want.

The effect would be an empty `${PREFIX}` output in case of an empty or undefined `${PREFIX}` input, while we wanted `/usr`.

This is also slightly related to why I prefer `${var+x}` instead of `$var`, it allows you to detect if the user or caller or whatever explicitly wanted an empty `var`, which would make no sense in the case of `PREFIX` or would it?! I guess `PREFIX=/` would be the same thing. But I find `+x` good practice alongside `set -u`.

I like the other suggestions, but maybe leave this one like so?

FFY00: You would be completely right if we didn't quote `${PREFIX+x}`. `-n` will evaluate to false when presented with an empty string, it will however evaluate to true when presented with, well, nothing. By quoting the variable we are creating a string, if the variable is empty the string will stay empty, but still a string. I think this can be better shown with examples:

`[ -n ]` evaluates to true
`[ -n "" ]` evaluates to false
`[ -n $test ]` evaluates to true when `$test` is empty
`[ -n "$test" ]` evaluates to false when `$test` is empty

We can drop the `+x` and have this being correctly evaluated with simpler code :grin:.

andrei-pavel:
Issues resolved through 9347910. Kind of took these amendments to other parts of the code as most of `data/scripts/install-man-pages.sh` was  copy-pasted from `install.sh` so I figured I'd fix them there as well. Code is consistent now in regards to style and quoting.

data/scripts/install-man-pages.sh
```diff
@@ -0,0 +1,24 @@
...
+fi
+export PREFIX
+
+echo "Installing man pages"
+install -Dm 644 'data/man/tilix' "$PREFIX/share/man/man1/tilix.1"
+gzip -f "$PREFIX/share/man/man1/tilix.1"
+
+if type po4a-translate >/dev/null 2>&1; then
+    for f in data/man/po/*.man.po; do
+        LOCALE="$(basename "$f" .man.po)"
```
FFY00: `$( ... )` statements don't need to be quoted, they already prevent word splitting.

install.sh
```diff
@@ -75,11 +78,10 @@ find po -name "*\.po" -printf "%f\\n" | sed "s/\.po//g" | sort > po/LINGUAS
 echo "Copying and installing localization files"
 for f in po/*.po; do
     echo "Processing $f"
-    LOCALE=$(basename "$f" .po)
-    msgfmt $f -o "${LOCALE}.mo"
-    install -d ${PREFIX}/share/locale/${LOCALE}/LC_MESSAGES
-    install -m 644 "${LOCALE}.mo" ${PREFIX}/share/locale/${LOCALE}/LC_MESSAGES/tilix.mo
-    rm -f "${LOCALE}.mo"
+    LOCALE="$(basename "$f" .po)"
```
FFY00: Same

install.sh
```diff
@@ -100,52 +102,34 @@ fi
...
-if type po4a-translate >/dev/null 2>&1; then
-    for f in data/man/po/*.man.po
-    do
-        LOCALE=$(basename "$f" .man.po)
-        install -d ${PREFIX}/share/man/${LOCALE}/man1
-        po4a-translate -k 0 -f man -m data/man/tilix -p data/man/po/${LOCALE}.man.po -l ${PREFIX}/share/man/${LOCALE}/man1/tilix.1
-        gzip -f ${PREFIX}/share/man/${LOCALE}/man1/tilix.1
-    done
-fi
+. "$(dirname "$(realpath "$0")")/data/scripts/install-man-pages.sh"
```
FFY00: Same.

data/scripts/install-man-pages.sh
andrei-pavel: Interesting, I didn't know that.

gnunn1:
Can I merge this now?

andrei-pavel:
> Can I merge this now?

Solved conflicts and removed meson.build which was unused after rebase. PR doesn't hold the original value after that, but I would still merge for consistent code style in the shell scripts. 👍

PR #1704, opened 2019-05-12
Meson: clean the build system fixes #1655

worldofpeace:
Same as #1655 but with the minor corrections needed.

@bilelmoussaoui noted
> LDC fails to link a lot of things here :/ Not sure if anyone can reproduce that

I can reproduce this in NixOS very easily with meson 0.49.2 (see https://github.com/gnunn1/tilix/commit/c4cf1fa019db78ca2bbb2fdaa48f54248b9bb16e).
However it does succeed to build for me with >=0.50.0 and not using the `release` buildtype.
We should probably report that as an upstream issue I think.

This technically conflicts with #1703 fixing the subdir placement, because it's fixed by using `@BASENAME@` here.


bilelmoussaoui:
Thanks for taking care of this!

worldofpeace:
@gnunn1 A review or merge here would be very helpful. Currently the meson build system here is broken.

gnunn1:
My meson knowledge is pretty much nil, I'll have a quick look and go ahead and merge unless there's a captain obvious problem.

worldofpeace:
> My meson knowledge is pretty much nil, I'll have a quick look and go ahead and merge unless there's a captain obvious problem.

I'd be happy to keep the `meson.build` things working here (probably others also) in that case.

gnunn1:
Thanks for the PR!

PR #1728, opened 2019-07-11
Use & when passing function references to HashTable constructor

myfreeweb:
ldc 1.16.0 complains about trying to call `g_str_hash` / `g_str_equal` with no arguments


source/gx/tilix/terminal/password.d
```diff
@@ -285,7 +285,7 @@ private:
 
     HashTable createHashTable() {
         import gtkc.glib;
-        return new HashTable(g_str_hash, g_str_equal);
+        return new HashTable(&g_str_hash, &g_str_equal);
```
FFY00: Use `ref` instead please.

https://en.wikibooks.org/wiki/D_(The_Programming_Language)/d2/Pointers,_Pass-By-Reference_and_Static_Arrays#Pass_By_Reference

myfreeweb: ref is a keyword in function signatures, not in calls??

FFY00: Yes, if you need this then do it in the function call. I assume this error will propagate to other calls, right? Or is it context specific? If then then this is fine.

myfreeweb: HashTable is obviously not a Tilix type, the constructor we're calling here is GtkD's wrapper around `g_hash_table_new`, we're giving it pointers to C functions (which `g_str_hash` and `_equal` are)

FFY00: So it seems this is the correct approach, I wish GtkD would provide more type safety.

myfreeweb:
Added patch for another problem with newer ldc (1.18.0):


```
/usr/local/include/d/object.d(3219,36): Error: cannot implicitly convert expression aa of type shared(ProcessStatus[int]) to const(shared(ProcessStatus)[int])
../source/gx/tilix/terminal/monitor.d(46,46): Error: template instance object.values!(shared(ProcessStatus[int]), shared(ProcessStatus), int) error instantiating
```

ximion:
From what it looks like, LDC builds Tilix just fine today, so I think GtkD added the necessary ref's to make this work.
However, the `shared` adjustment still seems relevant... If you have the time, could you please check if this is still needed and adjust the PR?

PR #1731, opened 2019-07-13
Fixes #1718 - Fix saving of already saved session

### LINKED ISSUE #1718: Session saving has no effect
Version de Tilix : 1.8.9
Version de VTE : 0.54
Version de GTK : 3.24.1

to reproduce : 
1) open a session using tilix -s /tmp/foo.json
2) in tilix GUI, "save"
-> /tmp/foo.json not modified
Expected : /tmp/foo.json updated

Workaround : make a "save as" and override the session file.

alexwhitman:
Previously the writing to a session file was only done if the Save As dialog was shown.  This PR fixes that and fixes #1718.

### DIFF
--- a/source/gx/tilix/appwindow.d
+++ b/source/gx/tilix/appwindow.d
@@ -1581,7 +1581,7 @@ private:
                         fcd.hide();
                         removeRecentSessionFile(fcd.getFilename());
                         error(e);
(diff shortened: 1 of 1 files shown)


gnunn1:
Thanks @alexwhitman

PR #1733, opened 2019-07-15
Avoid calling `values()` on a shared object

luk1337:
* This fixes a compilation on dmd v2.087.0+ ( I think? )
  /usr/include/dmd/druntime/import/object.d(3453,36): Error: cannot implicitly convert expression aa of type shared(ProcessStatus[int]) to const(shared(ProcessStatus)[int])
  source/gx/tilix/terminal/monitor.d(46,46): Error: template instance `object.values!(shared(ProcessStatus[int]), shared(ProcessStatus), int)` error instantiating

### DIFF
--- a/source/gx/tilix/terminal/monitor.d
+++ b/source/gx/tilix/terminal/monitor.d
@@ -43,7 +43,7 @@ private:
 
     bool fireEvents() {
         synchronized {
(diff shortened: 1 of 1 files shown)


luk1337:
Unsure whether this is correct as I never coded in D but it works here.

PR #1745, opened 2019-08-07
meson: add optional libsecret

worldofpeace:
This is optionally loaded with Linker.loadLibrary.


FFY00:
Can you elaborate a bit more in the commit message? Why is this needed? What will this enable? etc.

gnunn1:
I'm assuming this is for the password manager functionality in Tilix which relies on libsecret?

worldofpeace:
> I'm assuming this is for the password manager functionality in Tilix which relies on libsecret?

Correct, and we need meson to use pkgconfig to give the correct flags etc. or it isn't linked against it.

PR #1757, opened 2019-09-15
A few Meson improvements

ximion:
Tilix can actually be built with other D compilers than LDC now, so reflect that in the Meson file. Also, install the manual page (the untranslated only for now).


data/man/Makefile.am
```diff
@@ -3,15 +3,15 @@ manfile = $(builddir)/man/man1/tilix.1.gz
...
 
 $(builddir)/man/%/man1/tilix.1.gz: $(srcdir)/po/%.man.po
 	$(INSTALL) -d "$(builddir)"/man/$(*F)/man1; \
-	$(PO4A_TRANS) -k 0 -f man -m "$(srcdir)"/tilix -p $< -l "$(builddir)"/man/$(*F)/man1/tilix.1; \
-	gzip -f "$(builddir)"/man/$(*F)/man1/tilix.1; 
+	$(PO4A_TRANS) -k 0 -f man -m "$(srcdir)"/tilix.1 -p $< -l "$(builddir)"/man/$(*F)/man1/tilix.1; \
+	gzip -f "$(builddir)"/man/$(*F)/man1/tilix.1;
 
-EXTRA_DIST = $(srcdir)/tilix $(srcdir)/po/*.man.po 
+EXTRA_DIST = $(srcdir)/tilix.1 $(srcdir)/po/*.man.po
```
FFY00: Separate commit please.

ximion: Separate commit for what? This change is required for the manual page installation, which already is in a separate commit...

FFY00: My bad, I overlooked the git history when I pulled the changes.

Issue #1801, opened 2019-11-17
Add shortcut for unselect all (select none)


hannes-ucsc:
There is a shortcut for "Select all" so I think there should be a shortcut for the [inverse operation](https://developer.gnome.org/vte/unstable/VteTerminal.html#vte-terminal-unselect-all).

Tilix has this extremely useful feature that Ctrl-C only copies to clipboard if there is an active selection and otherwise sends the interrupt signal. I often find myself in the situation where I want to send the interrupt signal but have text selected. So I first need to deselect the text by using the mouse. It would be more ergonomic if I could deselect text via the keyboard.

alexwhitman:
Added in #1808

hannes-ucsc:
Confirmed working. Thank you, @alexwhitman & @gnunn1.

PR #1808, opened 2019-11-25
Add shortcut to "Unselect all"

alexwhitman:
This adds the option to bind a keyboard shortcut to "Unselect All", fixing #1801.  The state is set to `disabled` by default rather than trying to find an available and appropriate key binding.

This doesn't add a corresponding "Unselect All" option to the context menu which is easy enough to do but I couldn't find an inverse icon for `edit-select-all-symbolic`.


gnunn1:
Thanks!

PR #1810, opened 2019-12-03
add a color scheme inspired by PowerShell colors

druonysus:


druonysus:
any feedback would be welcome

gnunn1:
I'm not actively maintaining Tilix any longer, however in general I was not looking to add new themes directly into Tilix. There is a repo in github that contains a plethora of them and you can contribute your theme there.

PR #1822, opened 2020-01-07
Suppress deprecation messages

kubo39:


ximion:
This looks good to me, thank you for the patch! :-)

PR #1830, opened 2020-01-30
Avoid messing the previous exit code

edubxb:
The function ` __tilix_osc7()` in the `scripts/tilix_int.sh` script overrides the previous command exit code (when calling the`printf` command), so some tools that render custom prompts, like [Bash-it](https://github.com/Bash-it/bash-it), or [Starship](https://github.com/starship/starship) can't show the correct value.

This PR fixes it.

### DIFF
--- a/data/scripts/tilix_int.sh
+++ b/data/scripts/tilix_int.sh
@@ -43,7 +43,9 @@ __tilix_urlencode() (
 )
 
 __tilix_osc7() (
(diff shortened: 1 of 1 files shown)


edubxb:
Hello, any chance to get this PR merged? thanks!

ximion:
That makes sense, and the change looks good.
Thank you for your patience with this PR!

PR #1890, opened 2020-05-10
[scripts] fix `install.sh` script for directory errors

shinenelson:
I was hitting multiple directory creation errors when
attempting to install to a custom `PREFIX` earlier
today.

Upon further investigation, I found that 9347910e broke
the scripts when the `-d` flag for the `install`
command was replaced with the `-D` flag, but without
the right adaptation to the rest of the arguments.
The `install` command was stuck between 2 possible
options that arose from the change in flags.
See 6c1b0076 for more.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -49,7 +49,7 @@ echo "Installing to prefix $PREFIX"
 
 # Copy and compile schema
 echo "Copying and compiling schema..."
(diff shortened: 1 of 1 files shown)


ximion:
Indeed!
Thank you for the patch and sorry for the long delay in merging it.

shinenelson:
that's okay. everyone are only slowly getting back to normal life :)

PR #1927, opened 2020-07-27
[meson] add yaru scheme to install_data

### LINKED ISSUE #1926: Yaru scheme is not available
Yaru scheme, introduced with commit #1669, is not available in version 1.9.3.

To reproduce go to Preferences > Profiles Default > Color tab> Color scheme: Yaru is not available

It is probably missing from data/meson.build install_data, I am going to add a PR to fix it.

clobrano:
yaru.json is missing from data/meson.build and so it is not istalled.

Closes #1926


ximion:
Thank you for the patch! :-)

PR #1981, opened 2021-01-03
Update README.md

jeffdyke:
Shows a user how to ensure powerline fonts are configured correctly like the screenshots.  When looking for updated/improved emulators, this is always my first desire that powerline works out of the box...or close.  Thanks for the great project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,14 @@ Tilix requires the following libraries to be installed in order to run:
 * GSettings
 * [Nautilus-Python](https://wiki.gnome.org/Projects/NautilusPython) (Required for Nautilus integration)
 
+### Optional Fonts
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -60,6 +60,14 @@ Tilix requires the following libraries to be installed in order to run:
 * GSettings
 * [Nautilus-Python](https://wiki.gnome.org/Projects/NautilusPython) (Required for Nautilus integration)
 
+### Optional Fonts
+In the screeshots you'll notice I use powerline shell, in order to ensure this works correctly, you'll need to ensure the fonts are installed and Tilix is aware of your desired font:
```
ximion, line 64: Typo: "screeshots" --> "screenshots"

I would rather not speak with personal pronouns in a README, so maybe better use "In many of the screenshots, the `powerline` statusline shell plugin is used." etc.

jeffdyke: Good Point, addressed.  Thanks for considering this.

Issue #1658, opened 2019-02-28
Quake mode hides regardless of focus when 'Hide on focus lost' is On


anton-z-s:
Steps to reproduce:
1. Turn on 'Hide window when focus is lost'
2. Toggle quake mode by shortcut or `tilix -q` in another terminal, it doesn't matter
3. Quake mode is open and **get focus**, i can type in it
4. Terminal hides by itself after the time specified in 'Delay hiding' (even if it 5000)

It hides almost every time. 
Other Tilix settings or Gnome Shell extensions doesn't affect the issue. 
When 'Hide window when focus is lost' is off, the terminal doesn't hides by itself.
The issue doesn't reproduce under a new user. The same problem persists with other GTK drop-down terminals, like Tilda.
Clearly, there is a problem in my config, i would appreciate if you point me in the right direction.

I'm on Xorg, Ubuntu 18.10 (tested with both vanilla (without Canonical patches) and Ubuntu Gnome sessions).
```
Versions
	Tilix version: 1.8.1
	VTE version: 0.54
	GTK Version: 3.24.4
Tilix Special Features
	Notifications enabled=0
	Triggers enabled=0
	Badges enabled=1
```

PS. If i run Tilix from terminal, the only messages i get are:
```
(tilix:1587): Vte-WARNING **: 23:20:35.118: (../../src/vtegtk.cc:1954):int vte_terminal_match_add_gregex(VteTerminal*, GRegex*, GRegexMatchFlags): runtime check failed: (g_regex_get_compile_flags(gregex) & G_REGEX_MULTILINE)

(tilix:1587): Vte-WARNING **: 23:20:35.118: (../../src/vtegtk.cc:1954):int vte_terminal_match_add_gregex(VteTerminal*, GRegex*, GRegexMatchFlags): runtime check failed: (g_regex_get_compile_flags(gregex) & G_REGEX_MULTILINE)

(tilix:1587): Vte-WARNING **: 23:20:35.118: (../../src/vtegtk.cc:1954):int vte_terminal_match_add_gregex(VteTerminal*, GRegex*, GRegexMatchFlags): runtime check failed: (g_regex_get_compile_flags(gregex) & G_REGEX_MULTILINE)

[...]

Issue #1647, opened 2019-02-10
Short-term focus loss in a Quake mode


katoquro:
Hello!
I'm using [autokey](https://github.com/autokey/autokey) with tilix and when I hit an autokey shortcut it catches focus for a moment. With the option "Hide window when focus is lost" it is a bit inconvenient. 

If there were an option to set time interval within which returned focus doesn't considered as lost I could use tilix with autokey and autohiding feature.
Thanks!

gnunn1:
Isn't that what the existing "Delay hiding by (ms)" option already does in Quake options or am I mis-understanding?

katoquro:
Thanks for the quick reply!
I've tried setting it as max as possible (1000) and after my autokey macro done (I think less than 1 sec) tilix closes after that time. And I thought that option is about animation delay for the inevitable closing.
I use ubuntu 18.04 and I've tried updated to [1.8.9](https://launchpad.net/~webupd8team/+archive/ubuntu/terminix?field.series_filter=bionic) and have the same behaviour 

May be I can collect some debug info or event log?!

katoquro:
I'm not sure but can it be [here](https://github.com/gnunn1/tilix/blob/552923400bada2f0a1960e4b9a29b8588d46be31/source/gx/tilix/appwindow.d#L1752)?  
I mean additional check if window is not under the focus.

Btw is there a way to enable fine grained logging for debug?
Thanks!

gnunn1:
You can only enable fine-grained logging by building tilix from source code, i.e.

```dub build --config=trace```

And yes, you can try adding a check at that line to make sure the window doesn't have focus.

curiousercreative:
@katoquro did you have any success with this? I'm also using Autokey and suffering :/

katoquro:
@curiousercreative Yep. I've changed terminal to guake and added correspond functionality there on python )

Issue #858, opened 2017-03-28
quake mode pref `hide on focus lost` breaks expected hotkey behavior 


glitsj16:
Hi, I'm experiencing a problem with quake mode pref `hide on focus lost`. When I start tilix in quake mode after enabling the pref the tilix window appears with my defined layout. Pressing the quake mode hotkey hides the window, but it returns promptly on screen again. Disabling the `hide on focus lost` pref returns all things tilix back to normal.

Posting relevant info below. Anything out of the ordinary in my prefs that could cause this? Or am I not supposed to start quake mode with a session?

Regards.

= = =

$ tilix -q -s $HOME/.tilix-sessions/tilix.dropdown.json

= = =

$ cat $HOME/.tilix-sessions/tilix.dropdown.json
{
    "child": {
        "child1": {
            "child1": {
                "directory": "\/home\/glitsj16\/Downloads",
                "height": 400,
                "profile": "06f9ae90-dca8-4e14-8762-6cc23c707345",
                "readOnly": false,
                "synchronizedInput": true,
                "type": "Terminal",
                "width": 700
            },
            "child2": {
                "directory": "\/home\/glitsj16\/Downloads",
                "height": 400,
                "profile": "06f9ae90-dca8-4e14-8762-6cc23c707345",
                "readOnly": false,
                "synchronizedInput": true,
                "type": "Terminal",
                "width": 700
            },
            "orientation": 1,
            "position": 50,
            "type": "Paned"
        },
        "child2": {
            "directory": "\/home\/glitsj16\/Downloads",
            "height": 800,
            "profile": "06f9ae90-dca8-4e14-8762-6cc23c707345",
            "readOnly": false,
            "synchronizedInput": true,
            "type": "Terminal",
            "width": 700
        },
        "orientation": 0,
        "position": 50,
        "type": "Paned"
    },
    "height": 800,
    "name": "sysops",
    "synchronizedInput": false,
    "type": "Session",
    "version": "1.0",
    "width": 1400
}

= = =

$ tilix --version
Versions
	Tilix version: 1.5.4
	VTE version: 0.46
	GTK Version: 3.22.10

Tilix Special Features
	Notifications enabled=0
	Triggers enabled=0
	Badges enabled=0

= = =

$ dconf dump /com/gexperts/Tilix/ > tilix.dconf
$ cat tilix.dconf
[profiles]
[...]

PR #1993, opened 2021-01-23
keep quake window open if focus is restored before timeout has been reached

curiousercreative:
Simply allow the focusOut timeout (that closes the quake window) to be cancelled and specifically cancel it in the focusIn event handler.

Fixes:
- https://github.com/gnunn1/tilix/issues/1658
- https://github.com/gnunn1/tilix/issues/1647

Relates to:
- https://github.com/gnunn1/tilix/issues/858

### DIFF
--- a/source/gx/gtk/threads.d
+++ b/source/gx/gtk/threads.d
@@ -152,7 +152,7 @@ void threadsAddIdleDelegate(T, parameterTuple...)(T theDelegate, parameterTuple 
  *
  * This code from grestful (https://github.com/Gert-dev/grestful)
  */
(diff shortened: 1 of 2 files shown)


curiousercreative:
hmm, upon further testing this introduces a minor regression where quake window toggling can get funky

curiousercreative:
@ximion any chance this can be reviewed? I've been running it on both my desktop and laptop for many months (Pop!_OS 20.04, 20.10, 21.04). I'm reminded of this PR because I've just performed a system update that seems to have reinstalled a packaged version of this.

ximion:
The change looked good to me, thank you for the patch! :-)

PR #2006, opened 2021-02-24
Add meson target for man page translations

super7ramp:
Add a new meson custom which allows to generate and install man page
translations using `meson build` and `meson install` when po4a is
available.

This is a naive conversion of install-man-pages.sh.

### DIFF
--- a/data/man/meson.build
+++ b/data/man/meson.build
@@ -1,1 +1,25 @@
-install_man('tilix')
+install_man('tilix.1')
+
+po4a = find_program('po4a-translate', required: false)
+
+if po4a.found()
+
+    locales = [ 'ca', 'cs', 'de', 'en_GB', 'es', 'fr', 'hr',
+                'it', 'nb_NO', 'nl', 'oc', 'pl', 'pt_BR', 'pt',
+                'pt_PT', 'ro', 'ru', 'sr', 'tr', 'uk', 'zh_Hant' ]
+
+    foreach locale : locales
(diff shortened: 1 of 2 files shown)


ximion:
This makes sense to me, thank you for the patch!
It's odd though that Meson doesn't seem to have built-in functionality for this (at least I didn't find any information on that). Also, hardcoding language names sucks, but this is just as bad as the already existing shellscript, so, this is not worse at least :-P

data/man/meson.build
```diff
@@ -1 +1,25 @@
...
+        custom_target('man_' + locale,
+            output: locale,
+            input: [ 'tilix.1', 'po/' + locale + '.man.po' ],
+            command: [ po4a, '--keep', '0',
+                             '--format', 'man',
+                             '--master', '@INPUT0@',
+                             '--po', '@INPUT1@',
+                             '--localized', '@OUTPUT@/man1/tilix.1'],
+            install: true,
+            install_dir: get_option('mandir')
```
eli-schwartz, line 21: Needs to have `man1` specified at the end of the mandir, since the option is the base directory and `install_man()` adds a bit of magic on top of that to select the correct subdirectory (including for localized output filenames).

Unfortunately, install_man cannot accept as input, a built file from custom_target. And my efforts to add an API for custom_target functionality inside install_man, have stalled. :(

Oh well, one day I'll surely figure out something that makes everyone happy.

PR #2019, opened 2021-04-27
Disable advanced paste when there is no linebreak like iTerm2

kohnish:


kohnish:
closing this one. i accidentally made my source master, which isn't useful for me.
https://github.com/gnunn1/tilix/pull/2035

PR #2020, opened 2021-04-27
Dirty workaround for terminal window getting resized back to the orig…

kohnish:
…inal after advancedpaste box appears when it was fullscreen with quake mode.

### DIFF
--- a/source/gx/tilix/terminal/terminal.d
+++ b/source/gx/tilix/terminal/terminal.d
@@ -1438,10 +1438,15 @@ private:
         string pasteText = Clipboard.get(source).waitForText();
         if (pasteText.length == 0) return;
 
-        AdvancedPasteDialog dialog = new AdvancedPasteDialog(cast(Window) getToplevel(), pasteText, isPasteUnsafe(pasteText));
+        Window parent = cast(Window) getToplevel();
+        bool was_maximized = parent.isMaximized();
+        AdvancedPasteDialog dialog = new AdvancedPasteDialog(parent, pasteText, isPasteUnsafe(pasteText));
         scope(exit) {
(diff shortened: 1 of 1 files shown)


kohnish:
It doesn't work stably and came across the following documentation. It also might explain the original bug.
```
	/**
	 * Retrieves the current maximized state of @window.
	 *
	 * Note that since maximization is ultimately handled by the window
	 * manager and happens asynchronously to an application request, you
	 * shouldn’t assume the return value of this function changing
	 * immediately (or at all), as an effect of calling
	 * gtk_window_maximize() or gtk_window_unmaximize().
	 *
	 * Returns: whether the window has a maximized state.
	 *
	 * Since: 3.12
	 */
	public bool isMaximized()
	{
		return gtk_window_is_maximized(gtkWindow) != 0;
	}

	/**
	 * Asks to maximize @window, so that it becomes full-screen. Note that
	 * you shouldn’t assume the window is definitely maximized afterward,
	 * because other entities (e.g. the user or
	 * [window manager][gtk-X11-arch]) could unmaximize it
	 * again, and not all window managers support maximization. But
	 * normally the window will end up maximized. Just don’t write code
	 * that crashes if not.
	 *
	 * It’s permitted to call this function before showing a window,
	 * in which case the window will be maximized when it appears onscreen
	 * initially.
	 *
	 * You can track maximization via the “window-state-event” signal
	 * on #GtkWidget, or by listening to notifications on the
	 * #GtkWindow:is-maximized property.
	 */
	public void maximize()
	{
		gtk_window_maximize(gtkWindow);
	}
```

kohnish:
can't make it work stably, closing.

PR #2033, opened 2021-06-19
Add environment variable when it's quake mode

kohnish:
For users wanting to change the behaviour of shell rc based on mode.


ximion:
I see no reason not to have this feature, and the code looks fine to me. Thank you for the patch!

PR #2035, opened 2021-06-20
Disable advanced paste when there is no linebreak like iTerm2

kohnish:


ximion:
Makes sense, thank you for the patch!

PR #2058, opened 2021-10-20
Update flatpak manifest

amigadave:
Use the ldc and dmd extensions, rather than bootstrapping them. Update
the GNOME runtime to the (old and unsupported) 3.34, which is built on
the freedesktop 19.08 SDK, which is the newest version with the ldc and
dmd extensions. Add the necessary dependencies to build appstream for
using appstreamcli.


ximion:
Looks good to me, thank you for the patch!

PR #2062, opened 2021-11-10
Don't check GtkDragResult

luk1337:
Wayland does not set cancel reason thus this check always fails.

See: https://gitlab.gnome.org/GNOME/gtk/-/issues/2101


ximion:
Nice, thank you!

PR #2074, opened 2021-12-27
Make appstreamcli dependency optional

smortex:
Part of the code that use appstreamcli is conditionnaly executed
(validation) if the dependency is found, but another part is
unconditionnaly executed (generation).

Wrap all tthe code in the condition and mark the dependency as optional.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -35,25 +35,25 @@ endif
 
 # Merge release information into MetaInfo file
 tilix_metainfo_name = '@0@.appdata.xml'.format(project_id)
-appstreamcli = find_program('appstreamcli')
-metainfo_with_releases = custom_target('metainfo-news-merge',
+appstreamcli = find_program('appstreamcli', required: false)
+if appstreamcli.found()
+  metainfo_with_releases = custom_target('metainfo-news-merge',
     input : ['../NEWS', 'metainfo/@0@.in'.format(tilix_metainfo_name)],
     output : ['untranslated-@0@.appdata.xml'.format(project_id)],
     command : [appstreamcli, 'news-to-metainfo', '--limit=6', '@INPUT0@', '@INPUT1@', '@OUTPUT@']
-)
+  )
 
-# Install the MetaInfo file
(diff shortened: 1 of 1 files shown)


ximion:
Why? This will result in both a silently untranslated and uninstalled metainfo file, breaking discoverability of Tilix in software centers, and will break Flatpak builds due to missing release information in the (template) metainfo files.
If there is any merit to having this at all (as appstreamcli is fairly lightweight and available pretty much everywhere), putting this behind a default-on build option would be much better, so people need to consciously ignore it rather than getting a silent semi-broken build.

smortex:
> Why?

Mainly consistency: if some code checks whether the dependency was found, I assumed it was expected to be an optional dependency.

As a matter of fact, I was updating the FreeBSD port for tilix and appstreamcli is not available on this platform ATM (same for Flatpak), so the related tasks can be ignored on this platform as far as I understand.  When inspecting my build failure I saw I could skip this entirely and make the logic more consistent, hence this PR as it is now.

I have no objection to adjust this PR to add a meson option to opt-out of the appstreamcli stuff, how would you like me to name such an option @ximion?  `flatpak`?

```
option('flatpak', type : 'boolean', value : true)
```

eli-schwartz:
Perhaps it would be good enough, when appstreamcli is not found, to use `i18n.merge_file` on the metainfo file without first preprocessing it through appstreamcli in order to insert the `NEWS` contents?

It should still be discoverable by default, and translated as well... it would just not include release notes.

ximion:
The Flatpak SDKs do include `appstreamcli` & Co though (at least any newer ones) and Flatpak itself depends on it - so as far as Flatpak is concerned, this really shouldn't be an issue.

ximion:
This shouldn't be a problem any longer. If it still is for some reason, a new PR where the metainfo file is at least installed and translated without the NEWS entries would probably be okay as workaround (as long as these things are default-included). But `appstreamcli` is so ubiquitous that I doubt it's an issue to use it now.

Issue #2061, opened 2021-10-28
Meson 0.60.0 build error: Function does not take positional arguments


yochananmarqos:
https://mesonbuild.com/Release-notes-for-0-60-0.html

```
tilix/data/meson.build:13:0: ERROR: Function does not take positional arguments.
```

[meson-log.txt](https://pastebin.com/tWs6gETq)

PR #2081, opened 2022-01-23
Unbreak build with Meson 0.61.0

jbeich:
See https://github.com/mesonbuild/meson/issues/9441. Fixes https://github.com/gnunn1/tilix/issues/2061. Tested via 1.9.4 [downstream](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=260943).
```
$ meson setup _build
[...]
data/meson.build:13:0: ERROR: Function does not take positional arguments.
```
http://www.ipv6proxy.net/go.php?u=http://package18.nyi.freebsd.org/data/122amd64-default-foo/2022-01-20_18h39m40s/logs/errors/tilix-1.9.4.log

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -11,7 +11,6 @@ install_data(
 
 # Install the desktop file
 desktop_file = i18n.merge_file(
-    '@0@.desktop'.format(project_id),
     output: '@BASENAME@',
     input: 'pkg/desktop/@0@.desktop.in'.format(project_id),
     po_dir: meson.source_root() / 'po',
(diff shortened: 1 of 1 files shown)


jbeich:
See https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=260943 if you'd like to inspect tons of other similar fixes.

jbeich:
The fix builds fine downstream on FreeBSD after backporting on top of Tilix 1.9.4.

ximion:
Thank you for the patch! :-)

PR #2100, opened 2022-06-11
Update to GtkD 3.10.0

Jiehong:
I currently have a problem with building dmd on NixOS, so wasn't able to test it with dmd.

But building with ldc works fine.

The CI/CD should be running, though.

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -7,10 +7,10 @@
     "dflags-ldc": ["-disable-linker-strip-dead","-link-defaultlib-shared=false"],
     "dependencies": {
         "gtk-d:gtkd": {
-            "version": "3.9.0"
+            "version": "3.10.0"
         },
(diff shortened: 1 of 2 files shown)


Jiehong:
Using the version built with ldc locally seems to be working fine as far as I can tell :)

yochananmarqos:
The build fails with both `ldc` 1.29.0 and `dmd` 2.100.0.

Jiehong:
Anything blocking this PR?

yookoala:
@gnunn1 please kindly review this.

jbicha:
Debian Testing is using Tilix with gtk-d 3.10.0

ximion:
Yes, I think it's safe to bump requirements for people using dub as well :-)
Thanks for the patch!

PR #2101, opened 2022-06-11
bugfix: fix std.xml deprecation warning

Jiehong:
Fixing the only warning I see:

```
source/gx/tilix/prefeditor/prefdialog.d(960,16): Deprecation: module `std.xml` is deprecated - Will be removed from Phobos in 2.101.0. If you still need it, go to https://github.com/DigitalMars/undeaD
```

Everything seems to be the same other than that.

### DIFF
--- a/.github/ci/install-deps-deb.sh
+++ b/.github/ci/install-deps-deb.sh
@@ -28,4 +28,5 @@ eatmydata apt-get install -yq \
         libsecret-1-dev \
         libunwind-dev \
         libvted-3-dev \
+        libundead0 \
         po4a
--- a/dub.json
+++ b/dub.json
@@ -11,6 +11,9 @@
         },
         "gtk-d:vte": {
             "version": "3.9.0"
+        },
+        "undead": {
+            "version": "1.1.7"
         }
     },
     "buildTypes": {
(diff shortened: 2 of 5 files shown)


Jiehong:
The meson builds now fail. Any idea how to fix them? They say they can't read undead.xml now, but I've never used meson/ninja before.

yochananmarqos:
If this would merged, Tilix would have to depend on libundead. Not many distros package that.

Jiehong:
> If this was merged, Tilix would have to depend on libundead. Not many distros package that.

@yochananmarqos : you're right, it seems it's only present in Ubuntu/Debian repositories.

But it's missing from NixOS, Archlinux and Fedora.

Closing this then.

ximion:
You may be able to entirely replace the little XML parsing that Tilix does with GLib's own GMarkup parser: https://developer-old.gnome.org/glib/stable/glib-Simple-XML-Subset-Parser.html

yochananmarqos:
@Jiehong Well, Arch just did it anyway using the upstream undeaD tarball, see https://github.com/gnunn1/tilix/issues/2151#issuecomment-1439331926

PR #2111, opened 2022-08-11
Replace variables in custom link commands

### LINKED ISSUE #1820: Variables in custom links are not taken into account
Hi,

Context:
  - Tilix 1.9.1
  - Elementary OS 5.1

I use the custom links to open my IDE (vscode).
The problem I'm facing is that some tools are outputting relative paths and as far as I can see the command is executed from the user home directory, so my IDE is not finding the file...

I tried to workaround the issue by using the ${directory} as described in #1316 but if it dump the content of this variable, it's just empty...

Any help would be appreciated,

Thanks for this great tool !

arnested:
Fixes #1820.


ximion:
This looks okay, but I do not know this part of the code well enough to judge whether this would have unintended sideeffects. Some testing feedback would be very appreciated! :)

PR #2114, opened 2022-08-21
escape nautilus path

### LINKED ISSUE #2085: [Nautilus] bug: Tilix opens home directory instead of current directory if directory is in a SMB share.
Basically whenever I'm browsing a SMB share in Nautilus and want to open Tilix in my current directory, then pressing `Ctrl + Alt + T` opens the home directory.

This is a bit of a weird bug as pressing `Ctrl + Alt + T` opens the home directory in case described above, but right clicking the folder and pressing "Open Tilix Here" in the context menu opens the correct directory.

Stunkymonkey:
fixes #2085

### DIFF
--- a/data/nautilus/open-tilix.py
+++ b/data/nautilus/open-tilix.py
@@ -31,6 +31,8 @@ def _checkdecode(s):
 
 def open_terminal_in_file(filename):
     if filename:
+        # escape filename quotations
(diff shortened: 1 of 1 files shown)


Stunkymonkey:
@gnunn1 any chance this gets merged?

ximion:
This makes sense, thank you for the patch!

PR #2155, opened 2023-03-02
fix(security): avoid shell injection in open-tilix plugin

taoky:
I first reported this issue in https://bugs.archlinux.org/task/77698.

Currently tilix's open-tilix plugin for nautilus uses `subprocess.call()` with `shell=True`. However it fails to sanitize input data (filename) correctly, thus causing possible shell injection when filename contains `"` or `` ` ``, etc.

This PR tries to solve this issue by using `subprocess.Popen()` and avoiding invoking shell:

- [`Popen constructor`](https://docs.python.org/3/library/subprocess.html#popen-constructor) recommends to use `shutil.which()` to get actual path of executable (requires Python 3.3+).
- For `REMOTE_URI_SCHEME` part, [subprocess security considerations](https://docs.python.org/3/library/subprocess.html#security-considerations) recommends to use `shlex.quote()` to escape path (also requires Python 3.3+, and it may still have security issue when the shell is not POSIX-compliant).

### DIFF
--- a/data/nautilus/open-tilix.py
+++ b/data/nautilus/open-tilix.py
@@ -4,7 +4,9 @@
 # Shortcuts Provider was inspired by captain nemo extension
 
 from gettext import gettext, textdomain
-from subprocess import PIPE, call
+from subprocess import Popen
+import shutil
+import shlex
 try:
     from urllib import unquote
     from urlparse import urlparse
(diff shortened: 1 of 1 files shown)


ximion:
Can you adjust the patch so it works on top of the Nautilus plugin updates in master? Thanks for looking into this and creating this patch!

taoky:
> Can you adjust the patch so it works on top of the Nautilus plugin updates in master? Thanks for looking into this and creating this patch!

OK, the conflict has been resolved.

ximion:
Are you sure? It appears there's still a conflict and this PR can't be merged right now...

taoky:
> Are you sure? It appears there's still a conflict and this PR can't be merged right now...

This PR page shows "This branch has no conflicts with the base branch". I don't know where the "conflict" is.

image

And the "incomplete" label is incorrect - this PR is a complete work and is just waiting for reviewing & merging.

ximion:
Ah, the problem is that you have a merge commit in thie PR, so rebase does not work. But I can squash down the PR and merge it that way.

ximion:
Thank you for the patch! :-)

PR #2160, opened 2023-04-19
Fix compilation error due to std.xml deprecation

### LINKED ISSUE #2152: error during compiling tilix
HI
i always compile tilix successfully on my ubuntu 22.04, but now i have an error:

make: ***  Nessuna regola per generare l'obiettivo «clean».  Arresto.
    Starting Performing "release" build using /usr/bin/dmd for x86_64.
  Up-to-date gtk-d:gtkd 3.9.0: target for configuration [library] is up to date.
  Up-to-date gtk-d:vte 3.9.0: target for configuration [library] is up to date.
    Building tilix ~master: building configuration [default]
source/gx/tilix/prefeditor/prefdialog.d(960,16): Error: unable to read module `xml`
source/gx/tilix/prefeditor/prefdialog.d(960,16):        Expected 'std/xml.d' or 'std/xml/package.d' in one of the following import paths:
import path[0] = source/
import path[1] = /root/.dub/packages/gtk-d-3.9.0/gtk-d/generated/gtkd
import path[2] = /root/.dub/packages/gtk-d-3.9.0/gtk-d/generated/vte
import path[3] = /usr/include/dmd/phobos
import path[4] = /usr/include/dmd/druntime/import

Can you help me?

best regards

cifvts:
std.xml has been deprecated [1] and removed [2] from the standard library, therefor compilation failed due to the missing package.

As a temporary fix, add the undeaD package and use the xml library from it for the time being.

Fix #2152

[1] https://dlang.org/changelog/2.092.0.html#deprecate-xml
[2] https://dlang.org/changelog/2.101.0.html#remove-std-xml

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -11,6 +11,9 @@
         },
         "gtk-d:vte": {
             "version": "3.9.0"
+        },
+        "undead": {
+            "version": "1.1.8"
         }
     },
     "buildTypes": {
--- a/dub.selections.json
+++ b/dub.selections.json
@@ -1,6 +1,7 @@
 {
 	"fileVersion": 1,
 	"versions": {
-		"gtk-d": "3.9.0"
+		"gtk-d": "3.9.0",
+		"undead": "1.1.8"
 	}
 }
--- a/source/gx/tilix/prefeditor/prefdialog.d
+++ b/source/gx/tilix/prefeditor/prefdialog.d
@@ -957,7 +957,7 @@ private:
             return;
         }
 
-        import std.xml: DocumentParser, ElementParser, Element, XMLException;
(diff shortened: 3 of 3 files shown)


ximion:
This patch doesn't work for Meson (which in turn is what pretty much all Linux distros use to build Tilix), and relying on undeaD is, well, a dead end ;-) - and another dependency for people using Tilix to package.

I've updated the code to not rely on std.xml and use GMarkup instead.
Thanks for the patch anyway, it is appreciated!

cifvts:
Make sense, looking for another library was the next step, I will look what you have done, and use Meson myself, and see if the README could be updated too.
Thanks

PR #2162, opened 2023-04-23
Add close button to search box

### LINKED ISSUE #2131: The search bar can not be closed?
Control-F opens the search bar, but there is no close button.
I'm aware that you can press escape, but that is not easily disvoverable.

cifvts:
This patch add a close button to the search box, adding to pressing `esc` while on the search box input and the toggle button on the Window, to close the Search box area.

Fixes #2131

image


ximion:
That makes sense and the patch looks fine too!
Thank you!

PR #2165, opened 2023-05-02
Follow system color-scheme preference

flaktack:
On recent GNOME releases, it is possible to set a system-wide light/dark mode preference. tilix does not update its `theme-variant` based on this value. ([summary](https://blogs.gnome.org/alexm/2021/10/04/dark-style-preference/))

The preference is set in `org.gnome.desktop.interface` `color-scheme` and may be set to `default` / `prefer-light` / `prefer-dark`.

This PR modifies the existing `theme-variant` functionality so that
1. if `SETTINGS_THEME_VARIANT_KEY` is set to `system` (_default_) and a `color-scheme` value is set then a dark variant is preferred if `prefer-dark` is set.
2. whenever the system setting changes the `them-variant` preference is reapplied
3. `onThemChanged()` is emitted for them-variant change so that terminal colors may be reapplied if `SETTINGS_PROFILE_USE_THEME_COLORS_KEY` is set.

### DIFF
--- a/source/gx/tilix/application.d
+++ b/source/gx/tilix/application.d
@@ -103,6 +103,7 @@ private:
     enum MAX_BG_WIDTH = 3840;
     enum MAX_BG_HEIGHT = 2160;
 
+    GSettings gsDesktop;
     GSettings gsShortcuts;
     GSettings gsGeneral;
     GSettings gsProxy;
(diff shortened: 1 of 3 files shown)


ximion:
That makes sense, thank you for the patch!

PR #2179, opened 2023-07-29
Clarify tooltip of "reset to default" button for keyboard shortcuts

waldyrious:
This prevents ambiguity about whether the reset applies to the current shortcut or to all shortcuts

### DIFF
--- a/po/ak.po
+++ b/po/ak.po
@@ -1792,7 +1792,7 @@ msgid "Enable shortcuts"
 msgstr ""
 
 #: source/gx/tilix/prefeditor/prefdialog.d:805
-msgid "Set default"
+msgid "Restore default shortcut for this action"
 msgstr ""
 
 #: source/gx/tilix/prefeditor/prefdialog.d:916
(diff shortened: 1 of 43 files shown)


waldyrious:
Hey @ximion — sorry for the ping, but since you were the one who merged all of the [most recently merged PRs](https://github.com/gnunn1/tilix/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Amerged) (I checked at least 10), I wonder if you are still willing to review simple ones like this, or whether you've moved away from this project.

ximion:
Looks fine to me, except for maybe the coding style changes and l10n changes all bunched up in one commit - that's not great, but this change is small enough that it's not too much of an issue.
Thanks for the patch!

waldyrious:
Apologies — I would have been happy to split them up. I assumed it would make sense to make both changes as a single commit in the spirit of the "atomic commits" principle, which I interpret to mean that the codebase should be self-consistent at each snapshot (i.e. it would not make sense to make one change and not the other).

Anyway, I'll be mindful of the preference for separate i18n changes in future contributions :)

Thanks for merging!

ximion:
> Anyway, I'll be mindful of the preference for separate i18n changes (and code style changes, if substantial enough to stand alone) in future contributions :)

You can even completely ignore them - Tilix has a script to automatically extract them, which can be run at a later time in one commit (usually close before a release).

Thanks for contributing! Tilix does need all the help it can get ;-)

PR #2186, opened 2023-08-13
Replaced icon with more proportional and nice looking

Qupe:
New icon looks more proportional

Screenshot from 2023-08-13 17-04-00


waldyrious:
For future reference:

> more proportional and nice-looking

This essentially means making the icon square instead of rectangular — which I agree is a good move :)

> I will merge this and adjust it post-merge.

The adjustment was done in commit e0961a44c2ccc9e3b9289e070eb5348c487213be.

PR #2214, opened 2024-04-09
CI: use debian bookworm

kubo39:
libgtkd-3-0 package in debian trixie depends on libphobos2-ldc-shared105 and CI is failing, so uses debian bookworm.


kubo39:
Now libgtkd-3-0 package in debian trixie depends on libphobos2-ldc-shared106.

Issue #2210, opened 2024-02-22
Linking error with meson+dmd


rautyrauty:
when trying to build with meson+dmd, the following error occurs:
```
[137/171] Linking target tilix
FAILED: tilix 
[...]

PR #2219, opened 2024-04-29
meson.build: add `-allinst` when compiling with dmd or gdc, #2210

the-horo:
This is needed to work around a bug in dmd regarding template emissions.

Bug: https://github.com/gnunn1/tilix/issues/2210
Bug: https://issues.dlang.org/show_bug.cgi?id=20668

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -9,8 +9,13 @@ compiler = meson.get_compiler('d')
 if compiler.get_id() == 'llvm'
   d_extra_args = ['-vcolumns']
   d_link_args = []
-else
-  d_extra_args = []
+elif compiler.get_id() == 'dmd'
+  # Workaround for https://issues.dlang.org/show_bug.cgi?id=20668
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -9,8 +9,13 @@ compiler = meson.get_compiler('d')
 if compiler.get_id() == 'llvm'
   d_extra_args = ['-vcolumns']
```
taoky, line 10: The ldc compiler also needs `-allinst` to be set, or it would throw same error as others (as far as I could test with Arch Linux's `ldc`)

```suggestion
  d_extra_args = ['-vcolumns', '-allinst']
```

the-horo: It makes sense why we would add `-allinst` to all compilers given it's a frontend bug. I didn't add it for ldc2 since it was the only compiler that was working.

For the record, I tested with ldc2-1.39 and that one worked without the need of `-allinst`.

PR #2222, opened 2024-05-31
meson.build: Remove libunwind dependency

the-horo:
The dependency has been added in 6923ee9194523526a9d54d7be8ddcf51847b3c72 in the context of https://github.com/gnunn1/tilix/issues/1502#issuecomment-425723007 which was trying to diagnose issues with building using meson.

The commit has been added so others can use it as a starting point but it has never been removed. The dub configuration doesn't add it as a dependency as well.

Additionally update CI not to install it.


algj (approved):

ximion:
Makes sense, thanks!

Issue #2217, opened 2024-04-14
Right-Click to `clear` - Custom Commands in Roght-Click Context Menu


stellarpower:
Hi,

Thanks for Tilix, would be lost without it.

Something that exists in VS Code that I've become accustomed to is right-clicking and selecting clear to clear the output. It's useful for when I'm in the middle of a  long and verbose command, but also, is just faster than typing clear, and probably misspelling it three times or hitting caps lock instead of tab.

I miss it when I'm back on the root host and working in Tilix, so thought it might be a cool feature to add.

Beyond this, was wondering if it'd make sense to allow the user to customise the right-click menu, to be able to add any custom commands they want into a list there. A bit like we can do with most file managers, writing a little extension to add a custom menu entry. I guess clearing whilst halfway through a command wouldn't work if the menu simply runs that command in the shell, and there may be other commands that a user might like that would also require deeper integration into tilix itself, or some kind of logic, based on maybe the contents of the clipboard or what the user has just typed into the input buffer.

As a Qt user, I don't know very much about GTK, but from what I remember there's a similar meta-object system with GObject - so IDK if it'd be easy enough to expose some of the functionality built within Tilix into a minimal embedded scripting language or something, for a one-liner that could be executed. Similarly a bit like hos VS Code's launch JSON lets you add some variables in using dollar notation. I guess a trivial example could be something like `mkdir -p ${clipboardContents} ; chown root:root ${clipboardContents} ; chmod 0700 ${clipboardContents} `

But if that's a lot of work, just being able to run a custom command I think would still be very useful, even if that just works by calling into the shell with a string from the configs.

Cheers!

algj:
Vscode cleans terminal by removing all history except the current line. Currently what exists is `Right Click -> Other -> Reset and Clear`, which clears whole screen fully (default VTE behavior). But we need to keep the last line. This may need a hacky solution or a patch to VTE (which is the Virtual Terminal GTK widget library that Tilix is using).

Regarding about custom commands or right-click menu, I don't think anyone will develop such feature, the dev is currently looking for maintainers, you're free to look up the source code yourself :)

I'll take a look if it's possible anyway, since `Clear` menu item is indeed a useful thing to have.

PR #2232, opened 2024-07-28
Clear output action

algj:
Clear text while leaving current line intact, keeping selection and clearing scrollback buffer. Appears in shortcut list in `Terminal -> Clear output` and right-click menu as `Clear`.

Note: while this is a hacky method, it reproduces same "Clear" that vscode does, especially with TUI's.

#2217 :tada:


dejlek:
Isn't this what CTRL-L does in VTE?

PR #2248, opened 2024-12-30
Properly encode pasted text containing newlines

### LINKED ISSUE #1959: paste content in nano removes new lines
I have this strange behavior in nano using the Tilix terminal (works in another terminal) where, if I paste something, the new lines are completely removed, content is wrapped and new lines are created but based on the dimensions of the terminal window. `/bin/bash` is selected for both terminals and I don't know what to look for any more. 

For example when pasting:

    #!/bin/bash
    valid=true
    count=1
    while [ $valid ]
    do
    echo $count
    if [ $count -eq 5 ];
    then
    break
    fi
    ((count++))
    done

I get:

    !/bin/bash
    valid=true count=1 while [ $valid ] do echo $count if [ $count -eq 5 ]; then break fi ((count++))
    done

Tested both on manjaro and elementary os (ubuntu 18) and only tilix does this. Completely unusable since many of us rely on nano heavily.

akinomyoga:
This fixes #1959.

I received a report in my project (https://github.com/akinomyoga/ble.sh/issues/544).

The problem is that Tilix currently sends the raw text to the terminal application with its advanced paste dialog. However, the data sent to the terminal application is not plain text but should properly be encoded following the convention of the terminal communication. In particular, the newlines should be encoded by `\x0D` (which has an equivalent byte representation as CR `\r` in an ASCII string) in the traditional terminal protocol because `\x0A` (which has the same byte representation as LF `\n` in an ASCII string) means <kbd>C-j</kbd> in the convention. If one wants to send the multiline text preserving the different variations of newlines, when the terminal application requests the bracketed paste mode, the terminal should use the bracketed-paste-mode beginning/ending markers of the pasted text. All these processing is supposed to be handled by vte's `vte_terminal_paste_text`, but Tilix currently uses `vte_terminal_feed_child` to directly send the raw pasted text.

The same problem has been reported in #1959 for nano.

Ideally, we should use `vte_terminal_paste_text` through GtkD's `Terminal::pasteText`. However, this interface of GtkD has not yet been released. In this PR, we instead convert the newlines manually. I have confirmed that both issues #1959 and https://github.com/akinomyoga/ble.sh/issues/544 are fixed with this change.

### DIFF
--- a/source/gx/tilix/terminal/terminal.d
+++ b/source/gx/tilix/terminal/terminal.d
@@ -1447,7 +1447,7 @@ private:
         dialog.showAll();
         if (dialog.run() == ResponseType.APPLY) {
             pasteText = dialog.text;
-            vte.feedChild(pasteText[0 .. $]);
+            this.pasteText(pasteText[0 .. $]);
             if (gsProfile.getBoolean(SETTINGS_PROFILE_SCROLL_ON_INPUT_KEY)) {
                 scrollToBottom();
             }
@@ -1488,9 +1488,9 @@ private:
 
         if (gsSettings.getBoolean(SETTINGS_STRIP_FIRST_COMMENT_CHAR_ON_PASTE_KEY) && pasteText.length > 0 && (pasteText[0] == '#' || pasteText[0] == '$')) {
             pasteText = pasteText[1 .. $];
-            vte.feedChild(pasteText);
+            this.pasteText(pasteText);
         } else if (stripTrailingWhitespace) {
-            vte.feedChild(pasteText);
+            this.pasteText(pasteText);
         } else if (source == GDK_SELECTION_CLIPBOARD) {
             vte.pasteClipboard();
         } else {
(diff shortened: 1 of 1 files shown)


akinomyoga:
I requested a new release of GtkD (`gtk-d`) for `vte.Terminal.pasteText` in the upstream GtkD project, and they tagged a new release 3.11.0. I think it will be available on the DUB repository soon.

akinomyoga:
Since the latest version of GtkD (`gtk-d:3.11.0`) became available on the DUB repository, I updated the library version in commit 9635921342586c536879b7f67f3762f57216a3b6. Then, in commit 70781c8c4d6e9d41fe9c18b27e21b794beb696c9, I switched to using `vte.Terminal.pasteText` to send the pasted text, which is a dedicated method that VTE provides for pasting text.

akinomyoga:
@ximion I reverted the change to use the original commit for Gtkd-3.10.0 (where we implement `pasteText` by ourselves). I'll submit a version of using Gtkd-3.11.0's `pasteText` in a separate PR, so that it can be merged when Ubuntu LTS upgraded the Gtkd version.

@ximion Could you run the workflow with the current contents of the PR, which was adjusted for 3.10.0? Thank you.

ximion:
Much simpler - or so I thought...
All it needed was this: https://github.com/gnunn1/tilix/blob/master/.github/ci/make-install-deps-extern.sh

However, what I thought was just a simple change now yields new errors with GtkD 3.11: https://github.com/gnunn1/tilix/actions/runs/18931418631/job/54048906111

I guess I'll have to look into that tomorrow :-/

ximion:
Ah, easy fix then - please just rebase this PR on master :-)

ximion:
Oh man, GtkD upstream forgot to bump the version of its library! :-/
That's unfortunate! Can you, just for now, lower the dependency in the meson.build file, but keep all other changes? Then it should build (hopefully), and we need to fix that versioning issue upstream.

ximion:
Very good, thank you for the patch and sorry for the delays! :-)

akinomyoga:
Thanks!

PR #2279, opened 2025-12-09
Set prgname so that the app icon is correct to Wayland+GTK3

almson:


almson:
When are you going to make a new release for this and other changes?

PR #2298, opened 2026-08-23
Fix false-positive VTE configuration warning on early keypress

### LINKED ISSUE #982: VTE configuration warning pops up sometimes
I have this issue very long, but never reported it, as I have found no reliable way to reproduce it and it pops up very infrequently.

From time to time (very seldom) when I start Tilix the VTE configuration warning pops up. It happens mainly when I start Tilix completely new, but also when I just open another session or window. In general on my Arch system I have this configuration issue. I fixed it not by sourcing `vte.sh` but instead by using a login shell.

As this happens infrequently I suspect this is probably some kind of race condition somewhere.

loplex:
Pressing a key while the shell is still starting up (before its first prompt renders) locally echoes a character, moving the VTE cursor and firing `contents-changed`. Once the window title is later set (latching `terminalInitialized`), the very next `contents-changed` evaluation can find the directory still unreported — the OSC 7 report is typically sent a short but nonzero moment after the title, since common `vte.sh` integration scripts fork a helper process to URL-encode the cwd before sending it. This produces a false "VTE Configuration Issue" warning even though the shell is correctly configured.

Fixes #982.

Rather than deferring the check by a fixed delay and re-checking (which is either too short on a slow shell startup or needlessly long on a fast one), this records the cursor position the first time the condition looks suspicious and waits for it to change again before deciding. A further cursor move while the directory is still unset is a reliable sign of genuinely new shell output with no directory report ever coming, since the directory is always sent before the shell prints its visible prompt text. If the directory arrives instead, the pending state is cleared as soon as the outer condition stops holding.

The debounce decision itself is extracted into a small pure function (`checkVTEConfigWarning`) with a unit test, since it doesn't depend on a real VTE widget/pty.

Tested both with a unit test covering the debounce state machine directly, and manually in a real terminal session (by temporarily enabling the project's `USE_FILE_LOGGING` flag and reproducing with an early keypress right after opening a new terminal) — confirmed the false warning no longer fires and the underlying detection still works for shells that never report a directory.


source/gx/tilix/terminal/terminal.d
```diff
@@ -185,6 +185,77 @@ struct SyncInputEvent {
     string text;
 }
 
+/**
+ * Result of a single evaluation of the VTE configuration warning debounce, see #982 and
```
vazquezlopezeduardo53-afk, line 189: [BSCTrace.com 14/09/2026 00:01:32] I, hereby verify that I am the owner/creator of the address [0x55d398326f99059ff775485246999027b3197955]

vazquezlopezeduardo53-afk (approved):
Únete a MetaMask Rewards
https://link.metamask.io/rewards?referral=38657W

